Issue #81, opened 2023-03-21
New planner for Azure services


corinagum:
For customers to use non-public api too

PR #86, opened 2023-03-22
Added support for calling AzureOpenAI

### LINKED ISSUE #62: Change OpenAI wrapper -> axios
Adds flexibility to support more models

Stevenic:
- Created new axios based clients for calling OpenAI and AzureOpenAI.
- Created new AzureOpenAIPlanner class
- Removed dependency on openAI's library.

This update also includes base support for calling OpenAI's moderation api and creating embeddings.

Fixes #62 and #81

### DIFF
--- a/js/packages/botbuilder-m365/package.json
+++ b/js/packages/botbuilder-m365/package.json
@@ -29,10 +29,10 @@
         }
     },
     "dependencies": {
+        "axios": "^1.3.4",
         "botbuilder": "^4.19.0",
         "botbuilder-dialogs": "^4.19.0",
         "gpt-3-encoder": "^1.1.4",
-        "openai": "^3.2.1",
         "yaml": "^2.2.1"
     },
     "resolutions": {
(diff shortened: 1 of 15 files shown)


joe733:
Hi, how can I use [`google/flan-t5-small`](https://huggingface.co/google/flan-t5-small) model [served](https://github.com/SeldonIO/MLServer/) at `http://localhost:8080`, instead of OpenAI?

```ts
// Create AI components
const planner = new OpenAIPlanner({
  apiKey: config.openAIKey,
  defaultModel: "text-davinci-003",
  logRequests: true,
});
const moderator = new OpenAIModerator({
  apiKey: config.openAIKey,
  moderate: "both",
});
const promptManager = new DefaultPromptManager(
  path.join(__dirname, "./prompts")
);
```

I could only find this in the docs.

image

How do I do that in code?

vishn2200:
I have a similar problem

PR #135, opened 2023-05-19
Added support for authentication using the OAuthPrompt

Stevenic:
Added a new `Authentication` class that the `Application` class uses to automatically call the `OAuthPrompt` at the start of every turn.

The developer just needs to configure their Application instance with their OAuthPrompt settings.


rony2818:
Could you have sample for using the OAuthPrompt?
Because when i try to use config OAuthPrompt settings look like this:

```go
const app = new Application<ApplicationTurnState>({
    storage,
    ai: {
        planner,
        promptManager,
        prompt: 'chatGPT'
    },
    authentication: {
        connectionName: process.env.CONNECTION_NAME,
        text: 'Please login',
        title: 'Login',
        timeout: 30000 // User has 5 minutes to login.
    }
});
```
nothing happens after login
Screenshot 2023-06-30 at 16 25 55

Do I need to do anything more or just OAuthPrompt settings is enough?

singhk97:
@Stevenic when can we expect this module to be checked in?

corinagum (approved):

corinagum:
Closing in favor of #670, which is a branch off this PR and has more updates

PR #166, opened 2023-05-29
Add web.config to all samples to ensure Azure deployment works

Zenulous:
I recently tried to deploy some of these samples to Azure using Teams Toolkit. However, the apps do not start properly since the Bicep file creates Windows App Services, yet there is no `web.config` overrride to point the IIS Server to the right JS start file. I copied a sensible `web.config` which is also used by other Teams samples to ensure users do not have trouble deploying these samples to Azure.


corinagum:
@Zenulous thanks for the catch and contribution! I'm just double-checking with the Toolkit Team before merging your PR; we really appreciate the community participation.

corinagum (approved):

PR #190, opened 2023-06-22
bugfix: Add prompt revision in OpenAIModerator

singhk97:


corinagum:
Closed per scrum discussion

PR #215, opened 2023-06-29
feat: try add dotnet ci

swatDong:


swatDong:
Wrong base ref, close this.

Issue #208, opened 2023-06-27
[C# Port] AzureOpenAIModerator class


singhk97:
Implement the Azure Open AI moderator class just like the OpenAIModerator but instead uses AzureOpenAI content safety endpoint.

singhk97:
Status Update:
- `AzureOpenAIModerator` should be renamed to `AzureContentSafetyModerator`
- The API response is not the same as OpenAI's moderation API response. This should be handled accordingly instead of consolidating the two response.

PR #224, opened 2023-07-03
Add Azure Content Safety moderator

kuojianlu:
- add Azure Content Safety moderator to dotnet sdk
- add unit test for Azure Content Safety moderator
- add integration test for Azure Content Safety moderator
- update README of integration test

Resolves https://github.com/microsoft/teams-ai/issues/208

Notes:
- use REST API to call Azure Content Safety because it only has Python SDK
- place AzureContentSafetyClient into a separate folder because it is actually not part of OpenAI
- use `AzureContentSafetyTextAnalysisRequest(/Response)` rather than `AzureContentSafetyRequest(/Response)` because Azure Content Safety also handles images instead of only texts
- not map `AzureContentSafetyTextAnalysisResponse` to OpenAI's `ModerationResponse` because they are not that similiar and will lose some information like blocklistsMatchResults

### DIFF
--- a/dotnet/packages/Microsoft.Bot.Builder.M365/Microsoft.Bot.Builder.M365.Tests/IntegrationTests/AzureContentSafetyModeratorTests.cs
+++ b/dotnet/packages/Microsoft.Bot.Builder.M365/Microsoft.Bot.Builder.M365.Tests/IntegrationTests/AzureContentSafetyModeratorTests.cs
@@ -37,7 +37,7 @@ namespace Microsoft.Bot.Builder.M365.Tests.Integration
                 .Build();
         }
 
-        [Theory]
+        [Theory(Skip = "This test should only be run manually.")]
         [InlineData("I hate you", true)]
         [InlineData("Turn on the light", false)]
         [InlineData("我恨你", true)]
@@ -75,7 +75,7 @@ namespace Microsoft.Bot.Builder.M365.Tests.Integration
             }
         }
 
-        [Theory]
+        [Theory(Skip = "This test should only be run manually.")]
         [InlineData("I hate you", true)]
         [InlineData("The light is turned on", false)]
         public async Task AzureContentSafetyModerator_ReviewPlan(string response, bool flagged)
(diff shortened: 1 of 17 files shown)


dotnet/packages/Microsoft.Bot.Builder.M365/Microsoft.Bot.Builder.M365.Tests/IntegrationTests/AzureContentSafetyModeratorTests.cs
```diff
@@ -0,0 +1,110 @@
...
+                .AddUserSecrets<AzureOpenAICompletionTests>()
+                .Build();
+        }
+
+        [Theory]
+        [InlineData("I hate you", true)]
+        [InlineData("Turn on the light", false)]
+        [InlineData("我恨你", true)]
+        [InlineData("電気をつける", false)]
+        public async Task AzureContentSafetyModerator_ReviewPrompt(string input, bool flagged)
```
singhk97, line 45: These tests would fail if there is no working configuration (api key, endpoint) so we should make them manual tests by skipping them for now.

dotnet/packages/Microsoft.Bot.Builder.M365/Microsoft.Bot.Builder.M365.Tests/IntegrationTests/AzureContentSafetyModeratorTests.cs
```diff
@@ -0,0 +1,110 @@
...
+            else
+            {
+                Assert.Null(result);
+            }
+        }
+
+        [Theory]
+        [InlineData("I hate you", true)]
+        [InlineData("The light is turned on", false)]
+        public async Task AzureContentSafetyModerator_ReviewPlan(string response, bool flagged)
```
singhk97, line 81: We should make this a manual test and skip it for now.

singhk97 (changes requested):
PR Looks great! Only comment was to make the integration tests manual for now by skipping them. We shouldn't require devs to have access to the azure content safety resource just to have a passing test suite. We can come back to it later once we have the integration tests set up in the build pipeline.

dotnet/packages/Microsoft.Bot.Builder.M365/Microsoft.Bot.Builder.M365.Tests/IntegrationTests/AzureContentSafetyModeratorTests.cs
kuojianlu: Thanks for the notice. Skipped this test case.

dotnet/packages/Microsoft.Bot.Builder.M365/Microsoft.Bot.Builder.M365.Tests/IntegrationTests/AzureContentSafetyModeratorTests.cs
kuojianlu: Skipped this test case.

singhk97 (approved):

PR #275, opened 2023-07-13
[POC] Add Turborepo 

aacebo:
This is a POC on how using Vercels https://turbo.build/repo package might make the management of this monorepo more streamlined. 

Turborepo is a tool used to manage a monorepos build/test/lint pipelines and shared dependencies. Usage looks like:

`yarn install`
`npx turbo build`
`npx turbo lint`
`npx turbo clean`

I've also added some changes that would allow us to upgrade packages to their latest versions and unify dependency versions across the samples/lib


js/packages/teams-ai/src/AI.ts
```diff
@@ -18,7 +18,6 @@ import { TurnState } from './TurnState';
 
 /**
  * A function that can be used to select a prompt to use for the current turn.
- *
  * @template TState Type of the turn state.
```
corinagum, line 21: :o jsdocs should make up its mind. we just added these :/

corinagum:
@aacebo could you add a summary proposal as the top item in this thread? :)

js/packages/teams-ai/src/AI.ts
aacebo: ya I think this is because I was able to update alot of the package versions during this process 🤷‍♂️

corinagum: 🙃 Resolving - just me complaining

PR #286, opened 2023-07-17
fix custom turn state bug

singhk97:
If a custom turn state class was used, the temporary `input`, `history` & `output` values would not have been injected into the prompt template. This PR fixes that bug and adds a corresponding unit test


corinagum (approved):
LGTM

Issue #307, opened 2023-07-18
Initialize Project


aacebo:
Initialize python project using agreed upon tooling which should include:

- Build pipeline capabilities
- Automated test framework
- Lint framework
- Run Lint/Test on branch changes with GH merge protection
- Publish capabilities
- README documentation

aacebo:
https://github.com/microsoft/teams-ai/pull/320

PR #320, opened 2023-07-18
initialize python project with tooling

aacebo:
Initializing new `Python` project with tooling for testing/linting and some utility scripts.

https://github.com/microsoft/teams-ai/issues/307

This project is configured to format/lint to the PEP8 style guide

Some code was added as a way of testing that the lint/fmt rules were working correctly

Tools:

- `poetry`: like `yarn`, it makes project management easier via custom scripts and dependency management/virtual environments
- `pytest`: test lib like `jest` or `mocha`
- `pylint`: lint lib like `eslint`
- `yapf`: google's library used to format code, like `prettier`


blackchoey:
About branch policy, should we merge to `main` branch directly?

.github/workflows/python-ci.yml
```diff
@@ -0,0 +1,46 @@
+#
+# This workflow will build, run tests, and pack for the Microsoft.TeamsAI .NET SDK.
```
blackchoey, line 2: typo: should for the Python SDK

python/teamsai/prompts/prompt_manager.py
```diff
@@ -0,0 +1,69 @@
...
+import abc
+from typing import TypeVar, Generic, Any, Awaitable, Union
+from botbuilder.core import TurnContext
+from ..turn_state import TurnState
+from .prompt_template import PromptTemplate
+
+T = TypeVar("T", TurnState)
+
+
+class PromptManager(abc.ABC, Generic[T]):
```
blackchoey, line 15: We could use Protocol instead of ABC

blackchoey, line 15: We may use Protocol instead of ABC

.github/workflows/python-ci.yml
```diff
@@ -0,0 +1,46 @@
...
+        working-directory: ./python
+        run: |
+          python -m pip install --upgrade pip
+          pip install poetry
+          poetry install
+      - name: Check Dependencies
+        working-directory: ./python
+        run: |
+          poetry check
+      - name: Lint
```
blackchoey, line 35: Not sure whether we should split some checks into separate job, like Lint. So CI can list all the errors instead of stop at a certain step. You can make the decision.

python/teamsai/prompts/prompt_manager.py
aacebo: ok ya that makes sense to me I'll update it with my next PR that will be more focused on these pieces of code

.github/workflows/python-ci.yml
aacebo: I usually like to keep the commands separate to make the CI easier to read, ie a different command for each item, but still within the same job. I am open to creating different jobs for each tho so long as we have the .Net and JS projects do the same. Right now this matches what the C# team is doing.

aacebo:
> About branch policy, should we merge to `main` branch directly?

Yes I think ideally we won't publish this package until it is ready to be used but we should use Gitflow and incrementally merge to main instead of maintaining a giant feature branch. Any opposition to this approach @corinagum ?

blackchoey (approved):
Looks good to the CI part. Can merge as long as it's fine to merge into main branch based on your discussion with Kavin/Corina. If it's hard to make a quick decision, maybe we can create a `PYTHON` branch, just like the `DOTNET` one.

aacebo:
> Looks good to the CI part. Can merge as long as it's fine to merge into main branch based on your discussion with Kavin/Corina. If it's hard to make a quick decision, maybe we can create a `PYTHON` branch, just like the `DOTNET` one.

Ill put this in a `PYTHON` branch for now

PR #322, opened 2023-07-19
add step to publish pack to github nuget packages 

aacebo:
add a CI step to publish dotnet package to github so it can be used via nuget

### DIFF
--- a/dotnet/packages/Microsoft.TeamsAI/Microsoft.TeamsAI/Microsoft.TeamsAI.csproj
+++ b/dotnet/packages/Microsoft.TeamsAI/Microsoft.TeamsAI/Microsoft.TeamsAI.csproj
@@ -6,7 +6,7 @@
 	  <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
     <PackageId>Microsoft.TeamsAI</PackageId>
-    <Version>1.0.0</Version>
+    <Version>1.0.0-alpha0</Version>
     <Authors>Microsoft</Authors>
     <Company>Microsoft</Company>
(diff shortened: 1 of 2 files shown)


dotnet/packages/Microsoft.TeamsAI/Microsoft.TeamsAI/Microsoft.TeamsAI.csproj
```diff
@@ -1,10 +1,16 @@
...
   <PropertyGroup>
-	<TargetFramework>netstandard2.0</TargetFramework>
-	<LangVersion>latest</LangVersion>
-	<ImplicitUsings>enable</ImplicitUsings>
+	  <TargetFramework>netstandard2.0</TargetFramework>
+	  <LangVersion>latest</LangVersion>
+	  <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
+    <PackageId>Microsoft.TeamsAI</PackageId>
+    <Version>1.0.0</Version>
```
swatDong, line 9: Shall we use certain beta or preview version?

aacebo: that makes sense to me, do we have a convention for versioning we follow like semantic? cc @corinagum

aacebo: I think it makes sense to use `alpha` for private preview then `beta` for public preview

swatDong (approved):

PR #365, opened 2023-07-24
cleanup turn state and add tests

aacebo:
Cleanup turnstate code and add some unit tests.

PR #382, opened 2023-07-31
feat: turn state implementation and ut

blackchoey:
Continue work based on https://github.com/microsoft/teams-ai/pull/365. Thanks @aacebo .

Updated the original code to align with the js SDK behavior


aacebo (approved):
looks good, thanks for taking this over @blackchoey

blackchoey:
Finishes #385

Issue #395, opened 2023-08-02
AI moderator


blackchoey:

blackchoey:
Closing this as Alex is working on the moderator and have separate task to track.

Issue #386, opened 2023-08-01
AI planner


blackchoey:

corinagum:
Currently adding chat completion

Issue #206, opened 2023-06-27
Planner: update interface to remove completePrompt() method and change generatePlan() args


Stevenic:
As part of moving prompt management into the planner (#202) we will need to change the `Planner` interface some. We should remove `completePrompt()` all together leaving just `generatePlan()`. And for `generatePlan()` we should remove both the `prompt` and `options` args and instead just pass a pointer to the AI system. The final interface should look like this:

```TS
export interface Planner<TState extends TurnState> {
    /**
     * Generates a plan for the AI system to execute.
     * @param context Context for the current turn of conversation.
     * @param state Application state for the current turn of conversation.
     * @param ai AI system calling the planner.
     * @returns The plan that was generated.
     */
    generatePlan(
        context: TurnContext,
        state: TState,
        ai: AI<TState>
    ): Promise<Plan>;
}
```

govil1811:
P0

singhk97:
how are we expecting to specify the prompt when the `generatePlan` method does not accept one as input? Are we expecting it to be preconfigured within the planner itself?

Issue #203, opened 2023-06-27
Moderator: remove prompt from interface


Stevenic:
This is related to #202 where we're moving the prompt management to the planner. Since the moderator gets called before the planner there will be no prompt to pass in. The moderator should just moderate based off the users input which we can pass in.

govil1811:
P0

PR #398, opened 2023-08-02
feat: add interface for planner and moderator

blackchoey:
The new interface for planner and moderator after moves prompt management to planner. This is follow up of #202.

The `AI` class will call the planner and moderator in following ways:
``` python
async def chain(...):
    prompt_name = options.prompt # assumes user provides a prompt name to simplify the code
    user_input = turn_context.activity.text
    plan = await options.moderator.reviewPrompt(user_input)
    if plan is None:
        plan = await options.planner.generate_plan(turn_state, prompt_name)
        plan = await options.moderator.reviewPlan(plan)
    
    # process generated plan
```

The differences comparing with existing interface in js SDK:
1. For `Planner.geneate_plan`
    1. ~~Does not accept a `TurnContext` parameter any more, because this wasn't used in the implementation. I'm curious about the user scenario that requires `TurnContext`. Please add comments about the user scenarios and I'm happy to add it back.~~ Adds this back based on the potential scenario shared by Qianhao: it can be used to obtain Teams specific information when writing custom planner.
    2. The `prompt` parameter can now be either a prompt name, or a prompt template.
    3. The `options` parameter type is changed to `AIHistoryOptions` instead of a big `ConfiguredAIOptions`, because the current implementation only uses them. This is helpful for people to understand what options will take effect in the planner.
2. Removews `Planner.complete_prompt`, per #206 
3. For `Moderator.review_prompt`
    1. The `prompt` parameter is now just `user_input`, per #203 
    2. Does not accept `TurnContext`, `TurnState` and `ConfiguredAIOptions` parameter any more, because the moderator now can only reviews user input and seems no user scenarios that requires these parameter. Please add comments about the user scenarios and I'm happy to add them back.
4. For `Moderator.review_plan`
    1. Does not accept `TurnContext` and `TurnState` parameter any more. I'm curious about the user scenario that requires `TurnContext`. Please add comments about the user scenarios and I'm happy to add them back.
5. I also do not use generics for `TurnState` parameter in the interface, to reduce the need of type declaration when developers write code. Because planner and moderator is used by our code internally, developers actually do not care what's the type of `TurnState`. They just need to tell Application class the type of `TurnState` because they need to register handlers using the class and strong type makes the dev experience better. The updated coding experience to initialize Teams-AI classes would be:
    ``` python
    planner = OpenAIPlanner(...) # instead of: planner = OpenAIPlanner[ApplicationState](...)
    moderator = OpenAIModerator(...) # instead of: planner = OpenAIModerator[ApplicationState](...)
    app = Application[ApplicationState](...)
    ```

Related work items: #395 #386


blackchoey:
@Stevenic I saw your idea in #203 and #206 , but still trying to understand how they works with the AI class. While reviewing this PR, can you also share your thoughts behind the 2 issues? So I could see how to consolidate my PR with your thoughts.

python/teams_ai/planner/planner.py
```diff
@@ -0,0 +1,39 @@
...
+from typing import Any, Callable, Union
+from teams_ai.turn_state import TurnState
+
+class Planner(ABC):
+    """
+    interface implemented by all planners
+    """
+
+    @abstractclassmethod
+    async def generate_plan(self, state: TurnState, prompt_name_or_template: Union[str, PromptTemplate], *, history_options: AIHistoryOptions) -> Plan:
```
kuojianlu, line 16: I don't know what `turnContext` is for either. Guess it is extensible for the customized planner.

kuojianlu, line 16: How to get the prompt given the prompt name?

python/teams_ai/moderator/moderator.py
```diff
@@ -0,0 +1,38 @@
...
+    interface implemented by all moderators
+    """
+
+    @abstractclassmethod
+    async def review_prompt(input: str) -> Optional[Plan]:
+        """
+        Reviews user input before it's sent to the planner.
+
+        Args:
+            input (str): The user input.
```
kuojianlu, line 20: What Steven proposed seems to moderate the rendered prompt as well? Correct me if I misunderstand.

aacebo (commented):
is this meant to be a placeholder? I noticed it isn't exported from the `__init__.py` and it isn't properly referencing types `Plan` or `PromptTemplate`?

blackchoey:
> is this meant to be a placeholder? I noticed it isn't exported from the `__init__.py` and it isn't properly referencing types `Plan` or `PromptTemplate`?

Yes, I ignored the non-important parts so we can focus on the interface. I will declare the types later.

python/teams_ai/planner/planner.py
swatDong: Customized planner/moderator is one of the scenarios that may need `turnContext`.

IMO, `turnContext` is a good entry point to get teams specific context, e.g.,
- `turnContext.Activity.GetChannelData`
- `TeamsInfo.GetTeamDetailsAsync(turnContext, ...)`
- ...

It's easier to customize planner/moderator behavior per team if having `turnContext` as input.

blackchoey: The planner implementation will have its own prompt manager, it will take care of loading the prompt.

python/teams_ai/moderator/moderator.py
blackchoey: My understanding to #203 is since the planner implementation now uses a built-int prompt manager, there's no way to get the final prompt anymore. So the moderator can only review user's input.
But the interface doesn't care what's the content of input, maybe I can update the comment to avoid confusion.

python/teams_ai/planner/planner.py
blackchoey: @swatDong that makes sense to me, I will add `turnContext` as parameter

python/teams_ai/moderator/moderator.py
```diff
@@ -0,0 +1,38 @@
...
+from abc import ABC, abstractclassmethod
+from typing import Optional
+
+class Moderator(ABC):
+    """
+    interface implemented by all moderators
+    """
+
+    @abstractclassmethod
+    async def review_prompt(input: str) -> Optional[Plan]:
```
singhk97, line 15: Is there a reason we name this method `review_prompt` when we're actually reviewing the user input. Unless, is it the rendered prompt we are rendering?

corinagum: Adding comment for tracking :)

python/teams_ai/planner/planner.py
```diff
@@ -0,0 +1,40 @@
...
+from typing import Any, Callable, Union
+from teams_ai.turn_state import TurnState
+
+class Planner(ABC):
+    """
+    interface implemented by all planners
+    """
+
+    @abstractclassmethod
+    async def generate_plan(self, turn_context:TurnContext, state: TurnState, prompt_name_or_template: Union[str, PromptTemplate], *, history_options: AIHistoryOptions) -> Plan:
```
singhk97, line 16: according to #206, the parameters should be turn_context, state, & ai

python/teams_ai/planner/planner.py
```diff
@@ -0,0 +1,40 @@
...
+            prompt_name_or_template (Union[str, PromptTemplate]): The name of the prompt or a prompt template to use.
+            history_options (AIHistoryOptions): The options for the AI history.
+
+        Returns:
+            Plan: The generated plan.
+        """
+    
+
+    @abstractclassmethod
+    async def add_function(self, name: str, handler: Callable[[TurnContext, TurnState], Any], *, allow_overrides=False) -> None:
```
singhk97, line 32: is there a reason we have this here and not in the prompt manager class?

singhk97 (approved):
Other than a few comments, everything else looks good.

python/teams_ai/moderator/moderator.py
blackchoey: Do you have any ideas about the naming?
I didn't sync this with Steve. But since what users sent to OpenAI is called `prompt`, maybe we can keep using this name? The proposal #203 also keeps using `reviewPrompt` as function name for js SDK.

python/teams_ai/planner/planner.py
blackchoey: For `prompt_name_or_template`, I feel the prompt name / template is required, otherwise we won't know which prompt to use when generating the plan.
For `history_option`, it should be better to only accept options that is valid to the function. Otherwise, it could be hard to call this function because people don't know which options is required and need to initialize the whole AI instance, which is not necessary. Please let me know if there're other options available to the planner.

python/teams_ai/planner/planner.py
blackchoey: The goal of this PR is update existing interface to move prompt management to planner. My understanding is there should be no prompt manager concept exposed to developers in this case (though we indeed have a prompt manage implementation). So I move the `add_function` functionality to planner since it's necessary when rendering the prompt template.

PR #402, opened 2023-08-04
Stevenic/unit tests

corinagum:
Merge existing unit tests into `main` 
(code by @Stevenic )

### DIFF
--- a/js/packages/teams-ai/src/Application.spec.ts
+++ b/js/packages/teams-ai/src/Application.spec.ts
@@ -1,0 +1,147 @@
+import { strict as assert } from "assert";
+import { TestAdapter, MemoryStorage, ActivityTypes } from "botbuilder";
+import { Application } from "./Application";
+import { TestPlanner } from "./TestPlanner";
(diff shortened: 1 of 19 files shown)


js/packages/teams-ai/src/Application.spec.ts
```diff
@@ -0,0 +1,146 @@
+import { strict as assert } from 'assert';
```
blackchoey, line 1: should we put test related code to `test` folder?

corinagum: That's what Steve originally had, but then he decided to change it and removed the `test` folder. I'm not 100% sure why but I know it was intentional.

aacebo: I think for now we can merge this, we can always move tests to a `test` folder later on and its pretty standard practice in typescript projects for test files to live in the same directory as their corresponding source files.

aacebo (approved):

PR #412, opened 2023-08-09
Refactor Verify.ParamNotNull to use CallerArgumentExpression paramter attribute

singhk97:
`CallerArgumentExpression` is an attribute implemented in the compiler only for language version 10+ and is only available in .NET Core 3.0+ frameworks. Since we're targeting .NET standard 2.0, I added it as an internal to this library. Unit tests have been updated to reflect this.


aacebo (approved):

aacebo:
@singhk97 is this good to merge?

PR #413, opened 2023-08-09
Add Actions and Begin AI Class

aacebo:
adds boilerplate code for `AI` class and `actions`. Also cleans up some linting issues.

### DIFF
--- a/python/packages/ai/teams/ai/action/action_entry.py
+++ b/python/packages/ai/teams/ai/action/action_entry.py
@@ -1,0 +1,27 @@
+"""
+Copyright (c) Microsoft Corporation. All rights reserved.
+Licensed under the MIT License.
+"""
+
+from typing import Any, Awaitable, Callable, Generic, TypeVar, Union
+
+from botbuilder.core import TurnContext
+
+from teams.ai.turn_state import TurnState
+
+StateT = TypeVar("StateT", bound=TurnState)
+ReturnT = TypeVar("ReturnT")
+ActionFunctionSync = Callable[[TurnContext, TurnState, Any, str], Union[None, bool]]
+ActionFunctionAsync = Callable[[TurnContext, TurnState, Any, str], Awaitable[Union[None, bool]]]
+ActionFunction = Union[ActionFunctionSync, ActionFunctionAsync]
+
+
+class ActionEntry(Generic[StateT]):
+    name: str
+    allow_overrides: bool
+    func: ActionFunction
+
+    def __init__(self, name: str, allow_overrides: bool, func: ActionFunction) -> None:
+        self.name = name
+        self.allow_overrides = allow_overrides
+        self.func = func
--- a/python/packages/ai/teams/ai/action/default_action_types.py
+++ b/python/packages/ai/teams/ai/action/default_action_types.py
@@ -1,0 +1,16 @@
+"""
+Copyright (c) Microsoft Corporation. All rights reserved.
+Licensed under the MIT License.
+"""
+
+from enum import Enum
+
+
+class DefaultActionTypes(str, Enum):
+    UNKNOWN_ACTION = "__UnknownAction__"
+    FLAGGED_INPUT = "__FlaggedInput__"
+    FLAGGED_OUTPUT = "__FlaggedOutput__"
+    RATE_LIMITED = "__RateLimited__"
+    PLAN_READY = "__PlanReady__"
+    DO_COMMAND = "__DO__"
+    SAY_COMMAND = "__SAY__"
--- a/python/packages/ai/teams/ai/ai.py
+++ b/python/packages/ai/teams/ai/ai.py
@@ -1,0 +1,103 @@
+"""
+Copyright (c) Microsoft Corporation. All rights reserved.
+Licensed under the MIT License.
+"""
+
+from logging import Logger
+from typing import Dict, Optional
+
+from teams.ai.action import ActionEntry, ActionFunction, DefaultActionTypes
+from teams.ai.exceptions import AIException
+
+
+class AI:
+    """
+    ### AI System
+
+    The AI system is responsible for generating plans, moderating input and output, and
+    generating prompts. It can be used free standing or routed to by the Application object.
+    """
+
+    _log: Logger
+    _actions: Dict[str, ActionEntry] = {}
+
+    def __init__(self, log=Logger("default")) -> None:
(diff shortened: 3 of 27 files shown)


python/packages/ai/teams/ai/prompts/default_prompt_manager.py
```diff
@@ -20,8 +25,8 @@ class DefaultPromptManager:
     """
 
     _prompts_folder: str
-    _templates: dict[str, PromptTemplate]
-    _functions: dict[str, Callable[[TurnContext, TurnState], str]]
+    _templates: Dict[str, PromptTemplate] = {}
```
blackchoey, line 28: may remove the initialize logic in constructor?

python/packages/ai/teams/ai/prompts/default_prompt_manager.py
```diff
@@ -44,13 +49,13 @@ def add_function(
...
         if not allow_overrides and self._functions.get(name):
-            raise Exception(f"Function {name} already exists")
+            raise AIException(f"Function {name} already exists")
 
         self._functions[name] = handler
         return self
 
     async def render_prompt(
-        self, context: TurnContext, state: TurnState, name_or_template: str | PromptTemplate
+        self, context: TurnContext, state: TurnState, name_or_template: Union[str, PromptTemplate]
```
blackchoey, line 58: I'm fine with this change. But I remember `poetry run fmt` changes this to `str | PromptTemplate` for me. Just FYI. I will look at this further.

aacebo: ya I had to change this cause the `|` syntax was failing to compile

python/packages/ai/tests/prompts/test_assets/test_default_prompt_manager.py
```diff
@@ -39,32 +38,34 @@ async def test_render_prompt(self):
 
         # render prompt success when template is passed
         prompt_text = await prompt_manager.render_prompt(
-            mocked_turn_context, mocked_turn_state, "happy_path")
-        assert prompt_text == "test prompt"
+            mocked_turn_context, mocked_turn_state, "happy_path"
```
blackchoey, line 41: Are you using `poetry run fmt` for format the code? Seems my command gives different result, I'm wondering whether there's any local config we need to align?

aacebo: I switched us from `yapf` to `black` cause its more popular and consistent. I also changed both our tests to use `unittest` for assertions https://pypi.org/project/unittest/ cause its what the botbuilder stuff uses and it has alot of helper functions.

python/packages/ai/teams/ai/action/action_entry.py
```diff
@@ -0,0 +1,27 @@
...
+
+from typing import Any, Awaitable, Callable, Generic, TypeVar, Union
+
+from botbuilder.core import TurnContext
+
+from teams.ai.turn_state import TurnState
+
+StateT = TypeVar("StateT", bound=TurnState)
+ReturnT = TypeVar("ReturnT")
+ActionFunctionSync = Callable[[TurnContext, TurnState, Any, str], Union[None, bool]]
```
blackchoey, line 14: JS SDK requires the action to return a bool. Is there any reason we want to allow function return `None` here?

python/packages/ai/teams/ai/action/default_action_types.py
```diff
@@ -0,0 +1,16 @@
+"""
+Copyright (c) Microsoft Corporation. All rights reserved.
+Licensed under the MIT License.
+"""
+
+from enum import Enum
+
+
+class DefaultActionTypes(str, Enum):
+    UNKNOWN_ACTION = "__UnknownAction__"
```
blackchoey, line 10: JS SDK has **3** underscores in the name, should we align the name?

aacebo: the C# implementation supports `void` https://github.com/microsoft/teams-ai/blob/main/dotnet/packages/Microsoft.TeamsAI/Microsoft.TeamsAI/AI/Action/ActionHandler.cs#L28 and I think it makes sense to support and default to `false`

python/packages/ai/teams/ai/action/action_entry.py
aacebo: the C# implementation supports void https://github.com/microsoft/teams-ai/blob/main/dotnet/packages/Microsoft.TeamsAI/Microsoft.TeamsAI/AI/Action/ActionHandler.cs#L28 and I think it makes sense to support and default to false

python/packages/ai/teams/ai/ai.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def __init__(self, log=Logger("default")) -> None:
+        self._log = log
+        self._actions = {}
+        self.action(DefaultActionTypes.UNKNOWN_ACTION)(self.__unknown_action__)
+        self.action(DefaultActionTypes.FLAGGED_INPUT)(self.__flagged_input__)
+        self.action(DefaultActionTypes.FLAGGED_OUTPUT)(self.__flagged_output__)
+        self.action(DefaultActionTypes.RATE_LIMITED)(self.__rate_limited__)
+
+    def action(self, name: Optional[str] = None, allow_overrides=False):
```
blackchoey, line 32: JS SDK requires user to provide a name, may also mark the `name` parameter as required?

python/packages/ai/teams/ai/exceptions/ai.py
```diff
@@ -0,0 +1,9 @@
+"""
+Copyright (c) Microsoft Corporation. All rights reserved.
+Licensed under the MIT License.
+"""
+
+
+class AIException(Exception):
```
blackchoey, line 7: May rename to `AIError`, seems this is a common naming pattern.

python/packages/ai/teams/ai/action/action_entry.py
blackchoey: Can you create an issue so Corina can follow up in JS SDK (and can discuss there if there's difference opinion) ? I'm neutral to this and my only goal is keeping the functionality aligned across all 3 SDKs.

blackchoey (commented):
Approving. Left a few more comments to discuss.

blackchoey (approved):

PR #421, opened 2023-08-11
Added instructions for Azure Open AI to getting started

TheBluLion:


kevmcdonk:
It would be good to also add a note to say that the moderator should be disabled as well as don't believe that is supported in Azure Open AI currently.

TheBluLion:
Hi @kevmcdonk  Thanks for the feedback. I confirmed with the feature team that moderator for Azure OpenAI should be implemented and working now- give it a try and open an issue if you find that it is still not working.

PR #427, opened 2023-08-15
feat: AI planner - text completion

blackchoey:
1. Add AI planner implementation, currently only text completion is available.
1. Response parser to parse JSON plan and DO/SAY syntax in plain text
1. Do not use TypedDict in turn state data models so we can have better intellisense
1. DefaultPromptManager now takes our own `PromptTemplate` type as input, instead of semantick kernel's `PromptTemplate`
1. Update the lint command to only fail when there is fetal and error message. The warnings, refactors, etc. should be acceptable to some extent, for example TODO comments in our current stage.
1. Merge latest code from `PYTHON` branch

The code has been tested against the prompt in teams chef bot sample manually.

TODO: Add test cases for the code.


aacebo (approved):
@blackchoey looks great, thanks for working my changes into yours so quickly! Made a slight update to instead of ignoring warnings just exclude some rules that I don't  think are super important from the lintrc. In the future I think thats the preferred way to go so don't be afraid to do that in a PR next time if you think a pylint rule is just getting in the way and not adding value.

blackchoey:
> @blackchoey looks great, thanks for working my changes into yours so quickly! Made a slight update to instead of ignoring warnings just exclude some rules that I don't think are super important from the lintrc. In the future I think thats the preferred way to go so don't be afraid to do that in a PR next time if you think a pylint rule is just getting in the way and not adding value.

Thank you. As mentioned previously, I feel the warnings are valuable to see but low priority to fix, especially in the early stage. For example, the TODO comments. We can disable this first, but I think we need to invest more on the engineering system later.

PR #435, opened 2023-08-17
Tweak a few samples

stephentoub:
A little cleanup in ListBot sample, plus fix thread-unsafe use of Random in a few other samples

### DIFF
--- a/dotnet/samples/04.e.twentyQuestions/ResponseBuilder.cs
+++ b/dotnet/samples/04.e.twentyQuestions/ResponseBuilder.cs
@@ -2,8 +2,6 @@
 {
     public static class ResponseBuilder
     {
-        private static readonly Random RANDOM = new();
-
         private static readonly string[] SECRET_WORDS =
         {
             "Apple", " banana", "airplane", "coffee", "book", "ocean", "bird",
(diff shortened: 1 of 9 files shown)


dotnet/samples/04.e.twentyQuestions/ResponseBuilder.cs
```diff
@@ -141,10 +139,7 @@ public static string QuitGame(string? secretWord)
 
         public static string YouLose(string secretWord) => string.Format(GetRandomResponse(LOSE_MESSAGES), secretWord);
 
-        private static string GetRandomResponse(string[] responses)
-        {
-            int index = RANDOM.Next(responses.Length);
-            return responses[index];
-        }
+        private static string GetRandomResponse(string[] responses) =>
+            responses[Random.Shared.Next(responses.Length)];
```
corinagum, line 143: I'm not a `dotnet` dev; could you explain this one to me? :)

corinagum (approved):
One question, otherwise LGTM! We want at least one `dotnet` codeowner to approve before merging. I've assigned this internally. Thanks!

dotnet/samples/04.e.twentyQuestions/ResponseBuilder.cs
stephentoub: The previous version was using a `private static Random RANDOM = new Random()` instance. I don't know what the threading model is of these teams bots, but if there's any chance that multiple threads could concurrently end up calling this GetRandomResponse method, that would not be safe: like most .NET types, Random is not safe to be used by multiple threads concurrently and risks being corrupted with such use. Instead, the change switches to using the instance returned from `Random.Shared`, which _is_ thread-safe, so this is just switching from using a non-thread-safe instance to a thread-safe instance.  This line is saying "give me a random value in the range [0, response.Length), and return the value from responses at that index".

singhk97 (approved):

singhk97:
Thanks @stephentoub

Issue #432, opened 2023-08-16
Promp Moderator


aacebo:
Create moderator implementation that exposes an interface to validate prompts.

corinagum:
Evaluating OpenAI SDK for implementation

aacebo:
https://github.com/microsoft/teams-ai/pull/440

PR #440, opened 2023-08-17
Add Moderator Decorators & Chaining & OpenAIClient & ConversationHistory State

aacebo:
- add moderator decorators that can be used to implement customized moderation logic (spoke about creating a prototype that uses decorators instead of a moderator class with the team).
- add chaining to AI and Application classes
- add beginning of OpenAIClient http client (spoke about not using the official openai SDK with the team, we are not using it because it is inflexible and doesn't come with typings).
- add conversation history to conversation state (spoke with the team about creating a prototype for this instead of having a static class)

Resolves https://github.com/microsoft/teams-ai/issues/432


python/packages/ai/.pylintrc
```diff
@@ -188,6 +188,8 @@ function-naming-style=snake_case
 good-names=ai,
            ci,
            sk,
+           id,
+           n,
```
blackchoey, line 192: Seems `n` is not used as name in this PR?

python/packages/ai/teams/ai/clients/openai/schemas/moderation.py
```diff
@@ -0,0 +1,44 @@
...
+@dataclass
+class CreateModerationResponse:
+    @dataclass
+    class Result:
+        flagged: bool
+        categories: Dict[str, bool]
+        category_scores: Dict[str, int]
+
+        @classmethod
+        def from_json(cls, data: Dict[str, Any]) -> "CreateModerationResponse.Result":
```
blackchoey, line 19: May name the function as `from_dict`? This is how SK names the function, and `from_json` is used to initialize from json string.

python/packages/ai/teams/ai/clients/openai/schemas/moderation.py
```diff
@@ -0,0 +1,44 @@
...
+            return CreateModerationResponse.Result(
+                flagged=data["flagged"],
+                categories=data["categories"],
+                category_scores=data["category_scores"],
+            )
+
+        def to_dict(self) -> Dict[str, Any]:
```
blackchoey, line 26: The dataclass decorator should adds a function to return the dict. Could use that instead of defining our own function.

python/packages/ai/teams/ai/clients/openai/schemas/moderation.py
```diff
@@ -0,0 +1,44 @@
...
+    def from_json(cls, data: Dict[str, Any]) -> "CreateModerationResponse":
+        results: List[Any] = data["results"]
+
+        return CreateModerationResponse(
+            id=data["id"],
+            model=data["model"],
+            results=list(map(CreateModerationResponse.Result.from_json, results)),
+        )
+
+    def to_dict(self) -> Dict[str, Any]:
```
blackchoey, line 43: Same as previous comment, use the function added by dataclass decorator instead of writing our own

python/packages/ai/teams/ai/clients/openai/openai_client.py
```diff
@@ -0,0 +1,50 @@
...
+        )
+
+        if organization:
+            self.headers.add("OpenAI-Organization", organization)
+
+    async def create_moderation(
+        self, input: Union[str, List[str]], model: Optional[str] = None
+    ) -> OpenAIClientResponse[CreateModerationResponse]:
+        try:
+            res = await self.post("/v1/moderations", data={"input": input, "model": model})
```
blackchoey, line 42: If `model` is `None`, will OpenAI return error? Should we remove the `model` property in this case?

python/packages/ai/teams/ai/planner/azure_openai_planner_options.py
```diff
@@ -15,8 +15,8 @@ def __init__(
         self,
         api_key: str,
         default_model: str,
-        prompt_folder: str,
         endpoint: str,
+        prompt_folder: str = "prompts",
```
blackchoey, line 19: Why we need to give it a default value? The folder structure is totally defined by users so we may not make any assumptions to it.

python/packages/ai/teams/ai/planner/command_type.py
```diff
@@ -5,6 +5,6 @@
 from enum import Enum
 
 
-class CommandType(Enum):
+class CommandType(str, Enum):
```
blackchoey, line 8: Can you share why we need to inherit `str`?

python/packages/ai/teams/ai/planner/openai_planner.py
```diff
@@ -88,7 +84,11 @@ async def generate_plan(
             if assistant_prefix:
                 # The model sometimes predicts additional text
                 # for the human side of things so skip that.
-                position = result.lower().index(assistant_prefix.lower())
+                try:
+                    position = result.lower().index(assistant_prefix.lower())
+                except ValueError:
```
blackchoey, line 89: I have fixed this by using `find()` function in my PR, which is shorter than try/catch block. You can merge my changes after my PR is merged.

python/packages/ai/teams/ai/planner/openai_planner.py
```diff
@@ -159,9 +159,11 @@ async def _complete_text(self, prompt_template: PromptTemplate):
             prompt_template_config.completion
         )
 
-        text_completion_client = self._sk.get_ai_service(TextCompletionClientBase)(self)
+        text_completion_client = OpenAITextCompletion(
```
blackchoey, line 162: The `AzureOpenAIPlanner` may not work (I'm not sure), because it inherits `OpenAIPlanner`, and shared the logic here. And even it works, we may not expect SK ensures OpenAITextCompletion always work in the future.

python/packages/ai/teams/ai/planner/openai_planner_options.py
```diff
@@ -11,30 +11,9 @@
 class OpenAIPlannerOptions:
     api_key: str
     default_model: str
-    prompt_folder: str
+    prompt_folder: str = "prompts"
```
blackchoey, line 14: Also why we need to have default prompt_folder?

python/packages/ai/teams/ai/planner/openai_planner_options.py
```diff
@@ -11,30 +11,9 @@
...
     organization: Optional[str] = None
     endpoint: Optional[str] = None
     one_say_per_turn: bool = False
     use_system_message: bool = False
     log_requests: bool = False
```
blackchoey, line 19: The constructor is added on purpose to declare some parameter is keyword only, please do not remove it

python/packages/ai/teams/ai/planner/planner_error.py
```diff
@@ -0,0 +1,12 @@
+"""
+Copyright (c) Microsoft Corporation. All rights reserved.
+Licensed under the MIT License.
+"""
+
+from teams.ai.ai_error import AIError
+
+
+class PlannerError(AIError):
```
blackchoey, line 9: one guide line we want to follow is do not define new errors if developer cannot handle this error programatically. Can you share how do you expect users catch and handle this error?

python/packages/ai/teams/ai/prompts/prompt_manager.py
```diff
@@ -20,18 +20,19 @@
...
     Prompt manager used by action planner internally
     """
 
     _prompts_folder: str
     _templates: Dict[str, PromptTemplate] = {}
-    _functions: Dict[str, Callable[[TurnContext, TurnState], str]] = {}
+    _functions: Dict[str, Callable[[TurnContext, StateT], Awaitable[str]]] = {}
 
-    def __init__(self, prompts_folder: str) -> None:
+    def __init__(self, prompts_folder: str = "prompts") -> None:
```
blackchoey, line 35: Can you share why we need to have default value for prompts_folder?

python/packages/ai/teams/ai/ai_options.py
```diff
@@ -15,14 +17,26 @@
...
 @dataclass
 class AIOptions(Generic[StateT]):
+    planner: Planner
+    """
+    The planner to use for generating plans. For example,
+    you could set this as an instance of `OpenAIPlanner` or
+    `AzureOpenAIPlanner`.
+    """
+
+    prompt_manager: PromptManager[StateT] = PromptManager[StateT]()
```
blackchoey, line 27: Per steve's design, prompt manage is now part of planner and not exposed to user. Can remove this option.

python/packages/ai/teams/ai/ai_history_options.py
```diff
@@ -4,12 +4,9 @@
...
 
 from dataclasses import dataclass
-from enum import Enum
+from typing import Literal
 
-
-class AssistantHistoryType(str, Enum):
-    TEXT = "__TEXT__"
-    PLAN_OBJECT = "__PLAN_OBJECT__"
+AssistantHistoryType = Literal["text", "planObject"]
```
blackchoey, line 9: Can you share why we prefer this over an Enum?

python/packages/ai/teams/ai/prompts/prompt_manager.py
```diff
@@ -20,18 +20,19 @@
 
 SK_CONFIG_FILE_NAME = "config.json"
 SK_PROMPT_FILE_NAME = "skprompt.txt"
+StateT = TypeVar("StateT", bound=TurnState)
 
 
-class PromptManager:
+class PromptManager(Generic[StateT]):
```
blackchoey, line 26: Can you share why we need to use generic here? I may need an example to better understand.

python/packages/ai/teams/ai/ai.py
```diff
@@ -76,3 +100,221 @@ def __call__(func: Callable[[TurnContext, StateT, Any, str], Awaitable[bool]]):
...
+            Plan: Returns the same plan if the moderator approves the plan,
+            otherwise returns a new plan that redirects to a new set of actions.
+            Typically the moderator will return a new plan with a single DO command
+            that calls `AI.FlaggedOutputActionName` to flag the output for review.
+        """
+
+        self._review_plan = func
+        return func
+
+    def function(self, name: Optional[str] = None, allow_overrides=False):
```
blackchoey, line 165: Same with another comment, since prompt manager is part of planner, maybe we shouldn't add the entrypoint to prompt manager here? Just `planner.add_function()` is fine.

python/packages/ai/teams/ai/ai.py
```diff
@@ -76,3 +100,221 @@ def __call__(func: Callable[[TurnContext, StateT, Any, str], Awaitable[bool]]):
...
+                    )
+                )
+
+                state.conversation.value.history.add("assistant", text)
+            else:
+                state.conversation.value.history.add("assistant", plan.json())
+
+        for cmd in plan.commands:
+            if cmd.type == "DO":
+                do_cmd = PredictedDoCommand.from_dict(cmd.__dict__)
```
blackchoey, line 255: Maybe we can use `isinstance` function to detect the command type and finish the type inferring? The type of `plan.commands` can be changed to `Union[PredictedDoCommand, PredictedSayCommand]`. I personally not a fan of current method.

blackchoey:
@aacebo You mentioned a lot of new ideas in PR description, can you create new issues for them with some background (why we want this, etc.)

python/packages/ai/.pylintrc
```diff
@@ -439,7 +441,9 @@ disable=raw-checker-failed,
         R0903,
         W0212,
         W0511,
-        R1702
+        R1702,
+        W0622,
```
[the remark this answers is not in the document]
aacebo: This rule was for "too many branches" which doesn't seem super valuable to me right now since we are porting code that already exists. Later on if we want to remove these and refactor some things we can.

python/packages/ai/teams/ai/clients/openai/openai_client.py
aacebo: if it is done they default to gpt 3.5 I bellieve, it is an optional param in the other SDK's as well. https://github.com/microsoft/teams-ai/blob/main/dotnet/packages/Microsoft.TeamsAI/Microsoft.TeamsAI/AI/OpenAI/OpenAIClient.cs#L39

python/packages/ai/teams/ai/planner/command_type.py
aacebo: enums in python default to being `unknown`, if you want the enum to be of type string you need to also inherit from str

python/packages/ai/teams/ai/planner/azure_openai_planner_options.py
aacebo: This just provides a suggested default that we use in our samples, I think its useful to cut down on boilerplate as much as possible and devs can always pass in an override in they want to use there own structure. When possible I like to provide default input values that can act as useful examples or for devs that might just want a simple default implementation.

python/packages/ai/teams/ai/planner/openai_planner.py
aacebo: When running this code through real testing using the `ListBot`, this was throwing saying it could not find the service `TextCompletionClientBase` and that we were calling the method with too many params. I think something in `SK` isn't working for this method so I changed it to just use the class you are trying to get directly so that we can keep moving forward.

python/packages/ai/teams/ai/planner/openai_planner_options.py
aacebo: see above comment.

python/packages/ai/teams/ai/planner/planner_error.py
aacebo: Since our errors inherit they can handle by:

```python
@app.error
async def on_error(context: TurnContext, err: ApplicationError):
    if isinstance(err, PlannerError):
        .. handle planner error
        return
    .. handle normal error
```

python/packages/ai/teams/ai/prompts/prompt_manager.py
aacebo: see comment above about defaults

python/packages/ai/teams/ai/prompts/prompt_manager_error.py
```diff
@@ -3,7 +3,10 @@
 Licensed under the MIT License.
 """
 
+from teams.ai.ai_error import AIError
 
-class PromptManagerError(Exception):
-    def __init__(self, message: str) -> None:
-        super().__init__(message)
+
+class PromptManagerError(AIError):
```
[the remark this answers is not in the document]
aacebo: see comment above

python/packages/ai/teams/ai/state/message.py
```diff
@@ -0,0 +1,15 @@
+"""
+Copyright (c) Microsoft Corporation. All rights reserved.
+Licensed under the MIT License.
+"""
+
+from dataclasses import dataclass
+from typing import Literal
+
+MessageRole = Literal["system", "user", "assistant", "say", "do"]
```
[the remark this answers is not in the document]
aacebo: this was a suggestion made by Steve, instead of using the plain text conversation history implementation used in the other SDKs (which we will likely change soon) we are using a message object that we based off the one OpenAI uses https://platform.openai.com/docs/guides/gpt so we can more easily query history. DO and SAY are messages that may occur in the conversation history and we want the developers to have an interface for accessing them.

python/packages/ai/teams/ai/ai_options.py
aacebo: because the planner isn't using generics like the rest of the codebase it is throwing type errors so I can't use it for this until it is refactored to use a generic turn state 
image

python/packages/ai/teams/ai/ai_history_options.py
aacebo: String literals give you intellisense in the editor as you are working and do not require you to import a class/enum:
image

python/packages/ai/teams/ai/prompts/prompt_manager.py
aacebo: Like in the comment about the planner above, if the prompt manager doesn't use the same turn state as the ai/app class it will give type errors. If we want to remove generics we have to do it for the entire project so all the components are compatible. I will look into it today when I get some time.

python/packages/ai/teams/ai/ai.py
aacebo: I think since we have chosen to use decorators anything the user overrides should be accessible via a decorator. This provides a simple decorator for users to register functions the same way they would register an action, so the syntax is consistent.

python/packages/ai/teams/ai/ai.py
```diff
@@ -76,3 +100,221 @@ def __call__(func: Callable[[TurnContext, StateT, Any, str], Awaitable[bool]]):
...
+        """
+        Chains into another prompt and executes the plan that is returned.
+
+        This method is used to chain into another prompt. It will call the prompt manager to
+        get the plan for the prompt and then execute the plan. The return value indicates whether
+        that plan was completely executed or not, and can be used to make decisions about whether
+        the outer plan should continue executing.
+        """
+
+        rendered = await self._options.prompt_manager.render_prompt(context, state, prompt)
```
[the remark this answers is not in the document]
aacebo: see above comment about typing errors

python/packages/ai/teams/ai/ai.py
aacebo: good idea I didn't think of that 👍

python/packages/ai/teams/ai/planner/openai_planner_options.py
aacebo: dataclasses should not have constructors because they provide there own default constructor for the class. The way to initialize fields of a dataclass is via the method above using property initialization.

python/packages/ai/teams/ai/actions/action_entry.py
```diff
@@ -26,3 +26,8 @@ def __init__(
         self.name = name
         self.allow_overrides = allow_overrides
         self.func = func
+
+    async def invoke(
+        self, context: TurnContext, state: StateT, entities: Any, name: Optional[str] = None
```
corinagum, line 31: Is it possible to be more specific than `Any` here?

aacebo: the entities arg can be anything truly since its user defined, we could switch to `Dict[str, Any]` but thats essentially the same thing.

python/samples/echo/src/api.py
```diff
@@ -9,7 +9,7 @@
 from botbuilder.schema import Activity
```
corinagum, line 9: Please follow the samples naming conventions from `js` and `dotnet` (list bot too)

python/packages/ai/teams/app.py
```diff
@@ -3,12 +3,12 @@
 Licensed under the MIT License.
 """
 
-from typing import Awaitable, Callable, Dict, Generic, List, TypeVar
+from typing import Awaitable, Callable, Dict, Generic, List, Optional, TypeVar
 
-from botbuilder.core import Bot, TurnContext
-from botbuilder.core.activity_handler import ActivityTypes
+from botbuilder.core import Bot, BotFrameworkAdapter, TurnContext
```
corinagum, line 8: Does the python SDK not have `CloudAdapter`? It's the modern version of `BotFrameworkAdapter`

python/packages/ai/teams/ai/ai.py
```diff
@@ -76,3 +100,221 @@ def __call__(func: Callable[[TurnContext, StateT, Any, str], Awaitable[bool]]):
...
+        app.ai.review_prompt(on_review_prompt)
+        ```
+
+        Args:
+            context (TurnContext): The turn context.
+            state (StateT): The turn state.
+            prompt (PromptTemplate): The prompt.
+
+        Returns:
+            Optional[Plan]: Returns None if the moderator approves the input,
```
corinagum, line 126: Line character length seems kind of short to me. I'm inclined to recommend making it longer. `js` is 120, I believe.

python/packages/ai/teams/ai/actions/action_entry.py
corinagum: I'd prefer the later since that's how we do it in `js` but I'll defer to your experience with python

python/packages/ai/pyproject.toml
```diff
@@ -12,8 +12,8 @@ packages = [
 [tool.poetry.dependencies]
 python = ">=3.8,<4.0"
 botbuilder-core = "^4.14.4"
-aiohttp = "^3.8.5"
 semantic-kernel = "^0.3.7.dev0"
+aiohttp = "^3.8.5"
```
corinagum, line 16: (for my learning) why does this need to be moved?

corinagum (approved):
Some nits and questions, LGTM

python/packages/ai/pyproject.toml
aacebo: I was playing around with using the `openai` sdk so I removed it at some point but ended up readding it later on so it just got moved in that process.

python/packages/ai/teams/app.py
aacebo: ill look and see, all the python examples used this class so I didn't check if the cloud adapter was available. If it exists ill use that instead.

python/packages/ai/teams/ai/actions/action_entry.py
aacebo: I tried but for python that results in type errors since we what to pass class instances in some cases 
image

python/packages/ai/teams/app.py
```diff
@@ -144,6 +150,56 @@ async def on_after_turn(context: TurnContext, state: TurnState):
...
+        """
+
+        self._error = func
+
+        if self._adapter:
+            self._adapter.on_turn_error = func
+
+        return func
+
+    async def process_activity(self, activity: Activity, auth_header: str) -> InvokeResponse:
```
blackchoey, line 176: Seems this should be another feature that JS/C# should add. Can you help create new tasks for JS/C# SDK?

python/packages/ai/tests/prompts/test_prompt_manager.py
```diff
@@ -20,7 +20,7 @@
     UserState,
 )
 
-TEST_ASSERTS_FOLDER = "tests/prompts/test_assets"
+TEST_ASSERTS_FOLDER: str = "tests/prompts/test_assets"
```
blackchoey, line 23: Thanks for help adding this.
A minor question: I'm just curious about the practice in your mind: should we always add type annotation in our code? You could see some variable declarations does not add type annotation in our code, not sure whether we should update them all.

python/packages/ai/teams/ai/planner/azure_openai_planner_options.py
blackchoey: Thank you, does it make sense to make this parameter as keyword only parameter? If yes, can you move it after `*`?

python/packages/ai/teams/ai/planner/command_type.py
blackchoey: Thank you for the explanation. I'm curious why we want to make it to be type string? When we write `CommandType.Do`, do we need to care what's the type of the underlying value?

python/packages/ai/teams/ai/planner/openai_planner.py
blackchoey: See my comment at https://github.com/microsoft/teams-ai/pull/441#discussion_r1299504736. I would prefer to keep the implementation in my PR because that's how Semantic Kernel handles different (OpenAI & AzureOpenAI) client. Only use OpenAIClient here may cause potential issues when SK releases new package.

python/packages/ai/teams/ai/planner/openai_planner_options.py
blackchoey: Please also consider making it to be a keyword only parameter.

python/packages/ai/teams/ai/planner/planner_error.py
blackchoey: It also comes from Azure SDK guideline (which is summarize to the SDK guideline word I shared with you): do not create a new exception type unless the developer can handle the error programmatically.

This is a minor comment. If you think the guideline makes sense, let's adopt it. If not, we can go ahead with current implementation and discuss the guidelines further.

python/packages/ai/teams/ai/state/message.py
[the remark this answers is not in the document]
blackchoey: Agree that using the data models from OpenAI is better. But the document mentions the available roles are `system`, `user` and `assistant`. So seems `say` and `do` should not in the list?

python/packages/ai/teams/ai/ai_options.py
blackchoey: I see, you're declaring a single StateT instead of using 3 (ConversationT, UserT, TempT), seems mypy cannot handle this.
I'm good to go with current implementation. But I will create new PR to update planner as well as change the implementation to use planner. Let me know if you have any concerns to this.

python/packages/ai/teams/ai/ai_history_options.py
blackchoey: Do we need to align all Enums to Literal? How do we decide when to use Literal and when to use Enum?

python/packages/ai/teams/ai/prompts/prompt_manager.py
blackchoey: I'm good to go with your current implementation. But I wrote some examples here: https://github.com/microsoft/teams-ai/blob/1c886a283043b6bd2c1cee1455f08dfce3662c1a/python/packages/ai/examples/openai_planner.py#L45-L61. Not sure how you feels about it. It indeed introduces some difference with other SDK.

python/packages/ai/teams/ai/ai.py
blackchoey: I‘m neutral to this. Can you create tasks for JS/C# to add this? It should be doable in all SDKs and we can drive this if thinking it's valuable to users.

blackchoey:
Approving first. Please resolve the comments remained open before merge.

python/packages/ai/tests/prompts/test_prompt_manager.py
aacebo: that is a really good question that I have also thought about, We should decide together what practice is best for our team so that we are both on the same page. I _personally_ do the following: use explicit type annotations whenever exposing a type to the user ie options/class properties. I usually use inferred in cases like the one above where its internal only code, or if the language server for some reason gets it wrong which rarely but sometimes happens.

Lets discuss this in more detail during our weekly sync and if we decide to move towards using explict types everywhere I can open a PR to update the codebase and add linting to support it.

python/packages/ai/teams/ai/prompts/prompt_manager.py
aacebo: That makes sense to me, I'm going to create a task to handle that work separately but I will send it to you and start it today so that all our state uses that method.

python/packages/ai/teams/ai/ai_history_options.py
aacebo: My _personal opinion_ is that literals should be used for public types so that our users get intellisense and doesn't have to import another class from the library, thus making boilerplate smaller and easier to read. I am also for using literals everywhere in the codebase for this same reason, so if you think it makes sense to change all our enums to literal string types I support that and I can add it as a cleanup work item.

python/packages/ai/teams/ai/ai_options.py
aacebo: I think what you said below about using the three generics makes sense and if you are in aligment we can merge this PR as is for now and Ill open a new one to change all our `StateT` generics to use your method (3 generics) instead.

python/packages/ai/teams/ai/state/message.py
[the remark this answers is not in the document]
aacebo: Ya thats our own custom stuff because the `ConversationHistory` class needs to be able to do things like "get the last `say` message from history and replace it". I created this `Message` class to help do these queries in a more reliable way than using an array of strings like the JS implementation does because after talking with Steve he said we intend to make a similar change in the other libraries anyway. 

Do you have any ideas how we could do this better than what I've added in this PR? I am open to alternative approaches and maybe we can discuss this more in our weekly sync. If we come up with something better we can create a work item to improve upon or redo this history class entirely.

python/packages/ai/teams/ai/planner/planner_error.py
aacebo: Ok that works for me, In this case would we just have one error type `ApplicationError` or should we use a builtin error like `ValueError`? I am going to merge this today but during our sync we can discuss this more and create a work item to fix how we are handling errors in the codebase.

aacebo: When implementing this I was following the C# implementation but if the python way of doing things is to use builtin errors more then I'm fully behind doing it that way 👍

python/packages/ai/teams/ai/planner/openai_planner_options.py
aacebo: ok thanks for explaining that to me in more detail. It makes sense to me if we do what you said earlier actually and change these `Options` classes to just be constructor arguments instead, and the optional ones can just be positional args. Ill create a work item to track that work that will take place in a separate PR.

python/packages/ai/teams/ai/planner/command_type.py
aacebo: Ya its just good to know if an enum is a `string` enum or a `int` enum at least from an intellisense perspective. Not necessarily _needed_ since typing in python is optional anyway, but I personally like the types to be as accurate as possible especially in cases like these that don't require alot of work.

python/packages/ai/teams/ai/state/message.py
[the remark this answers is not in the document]
blackchoey: Thank you for the explanation. I fine with it now by knowing the reason of having "say" and "do".

PR #441, opened 2023-08-18
feat: AI planner - chat completion

### LINKED ISSUE #386: AI planner

blackchoey:
1. Support using chat completion API to generate plan.
1. Add `__history__` to ConversationState to support chat history\
1. Fix bug: when the prompt does not contain `assistant:` wording, the planner will throw error
1. Fix bug: calling OpenAI failed with argument error. This is a Semantic Kernel interface defect, where the interface definition is different than the actual class definition.

Resolves #386


python/packages/ai/teams/ai/planner/openai_planner.py
```diff
@@ -133,7 +140,9 @@ def _add_chat_completion_service(self) -> None:
 
     async def _complete_prompt(
         self,
+        turn_state: TurnState[ConversationT, UserT, TempT],
```
aacebo, line 143: this will probably break the typings since the ai class that uses it is using a generic type

python/packages/ai/teams/ai/planner/openai_planner.py
```diff
@@ -144,25 +153,71 @@ async def _complete_prompt(
...
+        state: TurnState,
+        user_message: str,
+        options: AIHistoryOptions,
+    ):
+        prompt_template_config = generate_sk_prompt_template_config(prompt_template)
+        request_settings = ChatRequestSettings.from_completion_config(
+            prompt_template_config.completion
+        )
+
+        chat_completion_client = self._sk.get_ai_service(ChatCompletionClientBase)(self._sk)
```
aacebo, line 182: this doesn't work, throws an error when run in a bot

python/packages/ai/teams/ai/planner/openai_planner.py
```diff
@@ -144,25 +153,71 @@ async def _complete_prompt(
...
+
+        chat_history: List[Tuple[str, str]] = []
```
aacebo, line 184: chat history is now in another class so please use that one, it exposes many helper methods that should make this work easier.

python/packages/ai/teams/ai/planner/openai_planner.py
```diff
@@ -144,25 +153,71 @@ async def _complete_prompt(
...
         prompt_template_config = generate_sk_prompt_template_config(prompt_template)
         request_settings = CompleteRequestSettings.from_completion_config(
             prompt_template_config.completion
         )
 
-        text_completion_client = self._sk.get_ai_service(TextCompletionClientBase)(self)
-        result = await text_completion_client.complete_async(
-            prompt_template.text, request_settings, NullLogger()
-        )
+        text_completion_client = self._sk.get_ai_service(TextCompletionClientBase)(self._sk)
```
aacebo, line 219: like above, this code doesn't work when run

python/packages/ai/teams/ai/state/conversation_state.py
```diff
@@ -12,3 +13,5 @@ class ConversationState:
     inherit a new interface from this base interface to strongly type
     the applications conversation state
     """
+
+    __history__: List[str]
```
aacebo, line 17: this was updated to be a class `ConversationHistory` which I spoke about with the team.

aacebo (changes requested):

python/packages/ai/teams/ai/planner/openai_planner.py
```diff
@@ -3,32 +3,39 @@
...
 from .planner import Planner
 from .response_parser import parse_response
 
+# Define role names used by OpenAI API
+SYSTEM_ROLE_NAME: str = "system"
+USER_ROLE_NAME: str = "user"
+ASSISTANT_ROLE_NAME: str = "assistant"
+
 
 class OpenAIPlanner(Planner):
```
aacebo, line 40: Also I noticed you logging requests using `print`, the `AI` class has a logger that can be passed in here for use. This is preferred because we want the user to be able to pass in their own logger instance if they want.

python/packages/ai/teams/ai/planner/openai_planner.py
```diff
@@ -144,25 +153,71 @@ async def _complete_prompt(
 
         result: str
```
aacebo, line 154: unecessary variable declaration

python/packages/ai/teams/ai/planner/openai_planner.py
```diff
@@ -144,25 +153,71 @@ async def _complete_prompt(
...
+            user_message = turn_state.temp.value.input
+            result = await self._complete_chat(
+                prompt_template, turn_state, user_message, history_options
+            )
+        else:
+            result = await self._complete_text(prompt_template)
 
         duration = datetime.now() - start_time
         # TODO: investigate how to get prompt/completion tokens
         self._log_request(f"\n{log_prefix} SUCCEEDED: duration={duration} response={result}")
```
aacebo, line 166: use logger instance here

python/packages/ai/teams/ai/planner/planner.py
```diff
@@ -10,10 +10,10 @@
 
 from botbuilder.core import TurnContext
 
+from teams import AIHistoryOptions
```
aacebo, line 13: this will result in a circular import

python/packages/ai/teams/ai/planner/openai_planner.py
blackchoey: The current code in this PR (which removes the extra logger parameter) works.

The type checking warning you mentioned in my previous PR should be ignored (and mypy in our CI does not complain this actually), because of a Semantic Kernel defect. If you look into the implementation, it actually doesn't implement the exact interface defined in `ChatCompletionClientBase` and caused the type checking error.

I have raised an issue to their repo and once they fixed the issue, the warning should be gone.

python/packages/ai/teams/ai/planner/openai_planner.py
blackchoey: See my previous comment about this. Resolving this to reduce concurrent discussions on same topic. We can discuss further in the other thread.

python/packages/ai/teams/ai/planner/openai_planner.py
blackchoey: created task for this https://github.com/microsoft/teams-ai/issues/444

python/packages/ai/teams/ai/state/conversation_state.py
blackchoey: See task #444

python/packages/ai/teams/ai/planner/openai_planner.py
blackchoey: Created task https://github.com/microsoft/teams-ai/issues/445

python/packages/ai/teams/ai/planner/openai_planner.py
blackchoey: Tracked with #445

python/packages/ai/teams/ai/planner/openai_planner.py
blackchoey: Thanks, removed

python/packages/ai/teams/ai/planner/planner.py
blackchoey: Can you share how circular import happens? pylint does not complains about circular import and I also haven't figured out the circular import manually.

But it indeed requires update, I will change it to `from teams.ai import AIHistoryOptions` instead of import from root module. This is a good catch. I missed this when using auto completion.

python/packages/ai/teams/ai/planner/openai_planner.py
blackchoey: I created `examples/openai_planner.py` to demonstrate the usage and also configured lint for the `examples` folder to ensure there's no typing issues. Please check that file and let me know what might break it.

python/packages/ai/teams/ai/planner/openai_planner.py
blackchoey: Since the code is both on active branch. If you don't mind and feels fine with my other comments, could merge my PR first and have additional commit to your PR to leverage ChatHistory class. I can help do this if you need.

python/packages/ai/teams/ai/planner/openai_planner.py
blackchoey: I see your problem with a new `StateT = TypeVar("StateT", bound=TurnState[ConversationState, UserState, TempState])` declared and used. I will investigate and create new PR for this as mentioned in your PR. Let's merge this PR first.

PR #469, opened 2023-08-24
[C#] Feat: #406 - Move from Newtonsoft.Json to System.Text.Json

### LINKED ISSUE #406: [C# Refactor] Replace Newtonsoft JSON with System.Text.Json in code base

singhk97:
* `ResponseParser` class is responsible for parsing plan json into a `Plan` object. It now uses System.Text.Json library for serializing & deserializing.

Note: The activity handler methods still use `Newtonsoft.Json` library as the Botframework SDK classes, such as `TeamsChannelAccount` are implemented to work only with Newtonsoft.Json.  When we unify these packages we should look into moving over to System.Text.Json then. 

Closes #406


aacebo (approved):

singhk97:
Verified that response parser works as expected but running the list bot sample.

corinagum:
Our requirements for new PRs will be changing, but we don't have the CI for it yet. 
Please modify the title to include Feat: #406 at the beginning.

kuojianlu (approved):

kuojianlu:
As Qianhao to review as I remember he has some comments on using Newtonsoft.Json or System.Text.Json.

dotnet/packages/Microsoft.TeamsAI/Microsoft.TeamsAI/AI/Planner/Plan.cs
```diff
@@ -1,28 +1,35 @@
...
+    /// </summary>
+    [JsonConverter(typeof(PlanJsonConverter))]
     public class Plan
     {
         /// <summary>
         /// A list of predicted commands that the AI system should execute.
         /// </summary>
-        [JsonProperty("commands")]
+        [JsonPropertyName("commands")]
+        [JsonIgnore]
```
swatDong, line 17: Will this totally ignore the `Commands` property?

dotnet/packages/Microsoft.TeamsAI/Microsoft.TeamsAI/AI/Planner/Plan.cs
```diff
@@ -33,8 +40,10 @@ public Plan(List<IPredictedCommand> commands)
         /// </summary>
         public string ToJsonString()
         {
-            // TODO: Optimize
-            return JsonConvert.SerializeObject(this);
+            return JsonSerializer.Serialize(this, typeof(Plan), new JsonSerializerOptions()
```
swatDong, line 43: Per https://github.com/microsoft/teams-ai/pull/462, is it better to make this singleton as well?

dotnet/packages/Microsoft.TeamsAI/Microsoft.TeamsAI/Utilities/JsonConverters/DictionaryJsonConverter.cs
```diff
@@ -0,0 +1,86 @@
+﻿using Newtonsoft.Json.Linq;
```
swatDong, line 1: Any compatible risk if mix-using `Newtonsoft.Json` and `System.Text.Json`? Or just a redundant namespace using?

dotnet/packages/Microsoft.TeamsAI/Microsoft.TeamsAI/Utilities/JsonConverters/PlanJsonConverter.cs
```diff
@@ -0,0 +1,120 @@
...
+
+        public override bool CanConvert(Type objectType)
+        {
+            return typeof(Plan).IsAssignableFrom(objectType);
+        }
+
+        public override Plan? Read(ref Utf8JsonReader reader, Type objectType, JsonSerializerOptions options)
+        {
+
+            if (reader.TokenType != JsonTokenType.StartObject)
```
swatDong, line 22: Any need to create `Utf8JsonReader readerClone = reader;` similar to the one in *CommandJsonConverter*?

Just curious, is it because `reader` is a ref and will be passed into child calls so need to create a clone? Correct me if I missed anything.

singhk97: So in this case we don't clone it because in the `JsonSerializer.Deserialize` method call, the reader's pointer is at the start of the command object (within the commands array). 

```
                            v----reader's pointer is here
{ type: 'plan', commands: [ {type: 'DO', action: 'GetItem'} ] }
```

In _CommandJsonConverter_, we want the reader's pointer to be at the start of the object. There's some minor duplication going on there, but I think in this case the code simplicity, trumps the performance hit.

dotnet/packages/Microsoft.TeamsAI/Microsoft.TeamsAI/Utilities/JsonConverters/DictionaryJsonConverter.cs
singhk97: it's redundant, removed it

dotnet/packages/Microsoft.TeamsAI/Microsoft.TeamsAI/AI/Planner/Plan.cs
singhk97: good call out

dotnet/packages/Microsoft.TeamsAI/Microsoft.TeamsAI/Utilities/JsonConverters/PlanJsonConverter.cs
swatDong: Thanks for the explanation. Yep agreed this is a good simple implementation.

swatDong (approved):

PR #487, opened 2023-08-26
feat #481 - fix warnings in the samples

### LINKED ISSUE #476: Teams Chef Bot sample clean up
In the TeamsChefBotActions sample, the first thing OnMessageActivityAsync does is check the turnContext.Activity.Type and compare it to ActivityTypes.Message. OnMessageActivityAsync can be invoked for activities other than messages? If so, that’s confusing. If not, the sample should be cleaned up to reflect that.

### LINKED ISSUE #481: Fix warnings in the samples
In the sample apps, either fix all warnings, or only enable the warnings / analyzers you care about. It’s confusing seeing a bunch of warnings in a sample app, which should be pristine. For example, the editorconfig is explicitly enabling CA2007 as a warning, but then the sample uses awaits without ConfigureAwait.

singhk97:
Cleaned warnings in the samples 

Most of the samples had the following warnings
* Add `ConfigureAwait` to async method call. I set the severity of this warning to _disabled_, since the guidance in application-level code to not include that.
* Asserting arguments like turn context & turn state of the OnMessageActivityAsync to be not null. It is pretty much guaranteed they won't be null as they are supplied by the SDK, so i set the severity of this warning to _disabled_ as well..
* Fix formatting..

Closes #481 
Closes #476


swatDong (approved):

kuojianlu (approved):

corinagum:
Hi @singhk97, my bad on the signed commits. I checked the branch history and incorrectly assumed it wouldn't cause issues for current team. Here's documentation if you need it: https://withblue.ink/2020/05/17/how-and-why-to-sign-git-commits.html

Issue #459, opened 2023-08-22
Consolidate AzureOpenAIPlanner and OpenAIPlanner into one


blackchoey:

blackchoey:
May be out of public preview's scope

blackchoey:
Wait for investigation result from js/C# sdk. Will port related changes to Python.

PR #508, opened 2023-08-31
[JS] refactor: #459 Refactor Planner

corinagum:
## Linked issues
#459 

NOTE: I have not yet done AzureOpenAIPlanner. I figure I should get feedback first. 

-----

## Main new concepts: `AIApiFactory` and `AIApiFactoryOptions` interfaces

`AIApiFactory` has two methods 
```ts
completePrompt(
    context: TurnContext,
    state: TState,
    prompt: PromptTemplate,
    options: ConfiguredAIOptions<TState>
): Promise<string>

generatePlan(
    context: TurnContext,
    state: TState,
    prompt: PromptTemplate,
    options: ConfiguredAIOptions<TState>
): Promise<Plan>
```

`AIApiFactoryOptions` has one property:
```ts
oneSayPerTurn?: boolean
```

The SDK will include two implementations of `AIApiFactory`, `OpenAIApiFactory` and `AzureOpenAIApiFactory`.

These two classes replace `OpenAIPlanner` and `AzureOpenAIPlanner`.

`OpenAIApiFactoryOptions` and `OpenAIApiFactoryOptions` will replace `OpenAIPlannerOptions` and `AzureOpenAIPlannerOptions`.

All properties on `OpenAIPlannerOptions` and `AzureOpenAIPlannerOptions` will be found in the new `AIApiFactoryOptions` interfaces. But since `oneSayPerTurn` is an SDK-level concept, it exists on the base interface, `AIApiFactoryOptions`.

## Using these provided `AIApiFactory` implementations:

(Name intentional; it won't be public) `PlannerImpl<Tstate>`, implements `Planner<Tstate>`. At construction time, PlannerImpl takes an AIApiFactory instance and keeps a reference to it at `PlannerImpl.aiApiFactory`.

When `PlannerImpl.completePrompt(...)` is called, it calls `this.aiApiFactory.completePrompt(...)`. It follows this same pattern for `generatePlan()`.

`PlannerImpl` will not be shipped as public API. Instead we will ship `createPlanner<TState>(aiApiFactory: AIApiFactory<TState>): Planner<TState>`

## Why `PlannerImpl` and `AIApiFactory`
`AIApiFactory` is the extension point for supporting any other LLM or multiple LLMs at once. These LLMs may be remote or run in-proc (!), or they may not use HTTP for communication.

The `PlannerImpl` is intentionally simple, its `AIApiFactory` does the heavy lifting.

## Developer-support for creating their own `AIApiFactory`
The helper function `createPlan()` is another new part of the public API. 

```ts
createPlan<TState extends TurnState = DefaultTurnState>(
    response: string,
    options: ConfiguredAIOptions<TState>,
    oneSayPerTurn: boolean = false
): Plan
```

`createPlan()` takes the response from the LLMs generated response and then applies the SDK concepts (`THEN DO`, `THEN SAY`) to return a properly formatted teams-ai `Plan`. The implementation details come from `OpenAIPlanner.generatePlan(...)`

> Note: `OpenAIApiFactory.generatePlan(...)` handles a 429 status code, if the factory wasn't rate limited, it passes the response to `createPlan()` and returns the result.


js/packages/teams-ai/src/OpenAIApiFactory.ts
```diff
@@ -0,0 +1,408 @@
...
+
+/**
+ * Prints the chat messages from a chat completion request.
+ * @param {ChatCompletionRequestMessage[]} messages The messages to print.
+ * @returns {string} The formatted chat messages.
+ * @private
+ */
+function printChatMessages(messages: ChatCompletionRequestMessage[]): string {
+    let text = '';
+    messages.forEach((msg) => {
```
aacebo, line 396: nit: use `for` loop instead of `forEach`

js/packages/teams-ai/src/PlannerImpl.ts
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * A planner that uses OpenAI's textCompletion and chatCompletion API's to generate plans.
+ * @summary
+ * This planner can be configured to use different models for different prompts. The prompts model
+ * will determine which API is used to generate the plan. Any model that starts with 'gpt-' will
+ * use the chatCompletion API, otherwise the textCompletion API will be used.
+ * @template TState Optional. Type of the applications turn state.
+ * @template TOptions Optional. Type of the planner options.
+ */
+class PlannerImpl<TState extends TurnState = DefaultTurnState> implements Planner<TState> {
```
aacebo, line 26: Whats the point of a distinction between `ApiFactory` and `PlannerImpl` if the planner is just a call through for the factory methods?

corinagum: `ApiFactory` is the main extension point for different LLMs.
based on a prompt, you might want to call a different service, or run locally. `PlannerImpl` is intentionally small so it will simply call the correct factory.

js/packages/teams-ai/src/AIApiFactory.ts
```diff
@@ -0,0 +1,97 @@
...
+    /**
+     * Optional. A flag indicating if the planner should only say one thing per turn.
+     * @summary
+     * The planner will attempt to combine multiple SAY commands into a single SAY command when true.
+     * Defaults to false.
+     */
+    oneSayPerTurn?: boolean;
+}
+
+export interface AIApiFactory<TState extends TurnState = DefaultTurnState> {
```
singhk97, line 27: This sounds like it should be named something like "LLMPlanner"

singhk97, line 27: This looks like it should be named something like "LLMPlanner"

js/packages/teams-ai/src/OpenAIApiFactory.ts
```diff
@@ -0,0 +1,407 @@
...
+            return result.data?.choices[0]?.text ?? '';
+        } else {
+            const temp = (state['temp']?.value ?? {}) as DefaultTempState;
+            const chatRequest = this.createChatCompletionRequest(state, prompt, temp.input, options);
+            const result = await this.createChatCompletion(chatRequest);
+            return result.data?.choices[0]?.message?.content ?? '';
+        }
+    }
+
+    public async generatePlan(
```
singhk97, line 113: Should this method be here? It seems like it's not directly related to orchestrating the AI service.

js/packages/teams-ai/src/AIApiFactory.ts
singhk97, line 27: Thoughts on naming this `LLMPlanner`, since it's identical to the planner interface, but specialized for LLMs i.e prompts

js/packages/teams-ai/src/AIApiFactory.ts
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ *
+ * @template TState
+ * @param {string} response The response from the OpenAI API.
+ * @param {ConfiguredAIOptions<TState>} options The options for the AI configuration.
+ * @param {boolean} oneSayPerTurn * Optional. A flag indicating if the planner should only say one thing per turn.
+ * @returns {Plan} The plan that was generated.
+ */
+export function createPlan<TState extends TurnState = DefaultTurnState>(
```
singhk97, line 51: It seems like the contents of this file related more to plans rather than the Open AI API itself, would it be better to rename this class?

corinagum: Not sure what you mean here. This isn't a class, just a helper function applicable to any service. The intent is that someone implementing their own factory could import this function and use it.
This file is not related to OpenAI, it's related to providing the Plan contracts to the SDK. Let me know if you have a different name suggestion based on the above.

js/packages/teams-ai/src/AIApiFactory.ts
corinagum: Are we planning on providing functionality outside of the chat domain that is (for example) `createImage` instead of `createPlan`? Unless we are, I'm not sure I understand the benefit of renaming this. Could you elaborate? Conceptually, LLMPlanner seems to apply more to SK than our library, since it implies the existence of other types of Planners.

js/packages/teams-ai/src/OpenAIApiFactory.ts
corinagum: `generatePlan` is making a call to the service, so imo it is related. Let me know if I'm missing something.

corinagum: `createPlan`, the private helper method, is the central SDK-specific code for making the `Plan`.  (`createPlan` is called inside `generatePlan`)

PR #523, opened 2023-09-07
[PY] feat: moderation/extensions part 1

aacebo:
## Linked issues

#minor

## Details

Provide a list of your changes here. If you are fixing a bug, please provide steps to reproduce the bug.

#### Change details

- move planner implementations to their own folder
- fix broken `OpenAIClient` implementation
- add moderator logic to `OpenAIPlanner`
- add `ActionTurnContext` to simplify action handlers

**code snippets**:

**screenshots**:

## Attestation Checklist

  - Local testing
  - E2E testing in Teams

### Additional information


python/packages/ai/teams/ai/planner/planner.py
```diff
@@ -58,3 +63,41 @@ def add_function(
             allow_overrides (bool, optional): Whether to allow overriding an existing function.
             Defaults to False.
         """
+
+    @abstractmethod
+    async def review_prompt(
```
blackchoey, line 68: The moderator functionality may be decoupled from planner like other SDK? As developers can choose a different moderator and even an empty moderator for their app.

python/packages/ai/teams/app.py
```diff
@@ -202,13 +205,97 @@ def __selector__(context: TurnContext):
...
             return False
 
-        def __call__(func: Callable[[TurnContext, StateT], Awaitable[bool]]):
+        def __call__(func: RouteHandler[StateT]):
+            self._routes.append(Route[StateT](__selector__, func))
+            return func
+
+        return __call__
+
+    def anonymous_query_link(self, command_id: Union[str, Pattern[str]]):
```
blackchoey, line 214: Do we need to add an extra `message_extension` layer for the functionalities like JS? So the code will be `app.message_extension.anonymous_query_link(...)`?

python/packages/ai/teams/ai/planner/planner.py
aacebo: No we are consolidating the moderator and planner since for each planner there is a cooresponding moderator

aacebo: this change will also be made in the other SDK's later on.

python/packages/ai/teams/app.py
aacebo: no that layer in JS doesn't do anything its just a class, I think its cleaner to have all our route selectors at the app level to keep it organized since the `message_extension` part isn't really doing anything.

python/packages/ai/teams/ai/models/openai/planner.py
```diff
@@ -115,10 +137,85 @@ async def generate_plan(
...
+                plan,
+            )
 
         return Plan()
 
+    async def review_prompt(
+        self,
+        _context: TurnContext,
+        state: TurnState,
+        _prompt: PromptTemplate,
```
blackchoey, line 152: do we need to review the prompt? Seems the parameter is not used here.

python/packages/ai/teams/ai/models/openai/planner.py
```diff
@@ -115,10 +137,85 @@ async def generate_plan(
...
+            res = await self._client.create_moderation(
+                state.temp.input,
+            )
+
+            if res.data.results[0].flagged:
+                return Plan(
+                    commands=[
+                        PredictedDoCommand(
+                            action=ActionTypes.FLAGGED_INPUT,
+                            entities=res.data.results[0].__dict__,
```
blackchoey, line 167: minor: `vars()` seems more preferred than `__dict__`, though they're doing the same thing. https://stackoverflow.com/questions/21297203/use-dict-or-vars

python/packages/ai/teams/ai/models/openai/planner.py
```diff
@@ -115,10 +137,85 @@ async def generate_plan(
...
+        self,
+        _context: TurnContext,
+        _state: TurnState,
+        plan: Plan,
+    ) -> Plan:
+        if self._options.moderate == "input":
+            return plan
+
+        for cmd in plan.commands:
+            if cmd.type == CommandType.SAY:
```
blackchoey, line 191: Use `isinstance` may be better. Example: https://github.com/microsoft/teams-ai/blob/PYTHON/python/packages/ai/teams/ai/ai.py#L259

python/packages/ai/teams/ai/prompts/prompt_manager.py
```diff
@@ -42,7 +42,7 @@ def __init__(self, prompts_folder: str = "prompts") -> None:
     def add_function(
         self,
         name: str,
-        handler: Callable[[TurnContext, TurnState], Awaitable[str]],
+        handler: Callable[[TurnContext, TurnState], Awaitable[Any]],
```
blackchoey, line 45: Is there any scenario that will return non-string value? The prompt template needs to replace corresponding variable with a string. There might be problem if developer does not return a string here?

python/packages/ai/teams/ai/ai.py
```diff
@@ -32,14 +42,11 @@ class AI(Generic[StateT]):
...
-    _review_prompt: Optional[
-        Callable[[TurnContext, StateT, PromptTemplate], Awaitable[Optional[Plan]]]
-    ] = None
-    _review_plan: Optional[Callable[[TurnContext, StateT, Plan], Awaitable[Plan]]] = None
 
-    def __init__(self, options: AIOptions, log=Logger("teams.ai")) -> None:
+    def __init__(self, options: AIOptions, *, log=Logger("teams.ai")) -> None:
         self._options = options
         self._log = log
+        self._options.planner.log = log
```
blackchoey, line 49: Maybe check whether the planner has a logger first to avoid overriding user defined logger?
By the way, should we just rely on developer to set the logger? If developers do not set a logger, it means they don't want logging. Current code seems will always add log for developers.

python/samples/light/src/bot.py
```diff
@@ -99,12 +102,10 @@ async def on_lights_off(
 
 @app.ai.action("Pause")
 async def on_pause(
-    context: TurnContext,
+    context: ActionTurnContext,
     _state: AppTurnState,
-    data: Any,
```
blackchoey, line 104: minor: I'm fine with the change. Can you share why you want to remove the data parameter and wrap it into ActionTurnContext? Want to be more clear about the reasons behind the senses.

python/packages/ai/teams/app.py
```diff
@@ -202,13 +205,97 @@ def __selector__(context: TurnContext):
...
+                context.activity.type != ActivityTypes.invoke
+                or context.activity.name != "composeExtension/submitAction"
+                or not context.activity.value
+                or not self._activity_with_command_id(context.activity, command_id)
+            ):
+                return False
+
+            message_preview_action = None
+
+            if isinstance(context.activity.value, object):
```
blackchoey, line 282: minor: the coding style is a bit different than `anonymous_query_link`. `anonymous_query_link` wraps the selector logic into a new function. You can see whether to unify the style.

python/samples/light/src/bot.py
aacebo: I talked it over with Steve / Kavin a bit last week, my reasoning is in the PR description, it is to simplify handler definitions for the developer and reduce dangling unused parameters. Since python flags the function as an invalid type if we don't include all the parameters, even if the developer doesn't want to use `data`, it might be a bit annoying.

If we wrap it in a new context specific to actions the handlers are simpler and the developer don't need to worry about the function signature as much since all our handlers will be essentially `async Function(TurnContext, State)`

python/packages/ai/teams/ai/ai.py
aacebo: I want the developer to be able to pass the logger into the `App` and have that logger passed down to all sub components for logging similar to how `slack` does it. This way if they want they can implement their own logger. I don't think the developer should have to pass the logger instance into every class below the app explicitly.

python/packages/ai/teams/app.py
aacebo: I do not see any difference, they are both using `__selector__` and wrapping in `__call__` the same exact way.

Issue #601, opened 2023-09-21
[C#] Implement AdaptiveCards fluent API


swatDong:
For example:

```csharp
// Create delegate to represent handler if needed
delegate Task<ActionExecuteResult> ActionExecuteHandler<TState>(ITurnContext turnContext, TState turnState, object data, CancellationToken cancellationToken)

// Methods ... should have overloads for different selector types
OnActionExecute(...)
OnActionSubmit(...)
OnSearch(...)
```

kuojianlu:
```csharp
// Action.Execute handler types
delegate Task<AdaptiveCard> ActionExecuteAdaptiveCardHandler<TState, TData>(ITurnContext turnContext, TState turnState, TData data, CancellationToken cancellationToken);
delegate Task<string> ActionExecuteTextHandler<TState, TData>(ITurnContext turnContext, TState turnState, TData data, CancellationToken cancellationToken);

// Action.Submit handler types
public delegate Task ActionSubmitHandler<TState, TData>(ITurnContext turnContext, TState turnState, TData data, CancellationToken cancellationToken);

// search handler types
class AdaptiveCardsSearchParams
{
    string QueryText { get; set; }
    string Dataset { get; set; }
}

class AdaptiveCardsSearchResult
{
    [JsonProperty("title")]
    string Title { get; set; }

    [JsonProperty("value")]
    string Value { get; set; }
}

delegate Task<IList<AdaptiveCardsSearchResult>> SearchHandler<TState>(ITurnContext turnContext, TState turnState, Query<AdaptiveCardsSearchParams> query, CancellationToken cancellationToken);

// AdaptiveCards
class AdaptiveCards<TState, TTurnStateManager>
    where TState : ITurnState<StateBase, StateBase, TempState>
    where TTurnStateManager : ITurnStateManager<TState>, new()
{
    AdaptiveCards(Application<TState, TTurnStateManager> app)

    Application<TState, TTurnStateManager> OnActionExecute<TData>(string verb, ActionExecuteAdaptiveCardHandler<TState, TData> handler)
    Application<TState, TTurnStateManager> OnActionExecute<TData>(Regex verbPattern, ActionExecuteAdaptiveCardHandler<TState, TData> handler)
    Application<TState, TTurnStateManager> OnActionExecute<TData>(RouteSelector routeSelector, ActionExecuteAdaptiveCardHandler<TState, TData> handler)
[...]

PR #664, opened 2023-10-10
[C#] fix: use string.Equals instead of Regex

kuojianlu:
closes: #601


dotnet/packages/Microsoft.TeamsAI/Microsoft.TeamsAI/Application/AdaptiveCards.cs
```diff
@@ -148,7 +148,7 @@ public AdaptiveCards(Application<TState, TTurnStateManager> app)
         /// <returns>The application for chaining purposes.</returns>
         public Application<TState, TTurnStateManager> OnActionExecute(string verb, ActionExecuteAdaptiveCardHandler<TState> handler)
         {
-            return OnActionExecute(new Regex($"^{verb}$"), handler);
```
swatDong, line 151: how about just using `string.Equal()`?
does `Escape` correctly handle line breakers without `Singleline` or `Multiline` option?

kuojianlu: Oh I see, this is not due to `Escape` but because `$` allows a trailing newline `\n` character...
You're right, regex is error-prone, let me use `string.Equal` instead.

PR #670, opened 2023-10-12
[JS] feat: Authentication

singhk97:
## Linked issues

closes: #738, #739
## Details

Foundations of handling user authentication.

1. `Authentication` class encapsulates logic to authentication for a single connection.
   - It currently supports bot and messaging extension authentication through the OAuthPrompt, using BF Token service implicitly.
   - Subsequent PR will include support bot and messaging extension authentication without BF Token service.
1. `AuthenticationManager` class handles multiple user configured connections and exposes `signInUser` and `signOutUser` methods for application code initiate authentication.
1. Users create `AuthenticationOptions`  object to configure authentication:
   - `settings`: contains one or more user authentication settings
   - `default`: the default setting to use in the auto sign in flow.
   - `autoSignIn`: whether to sign user in by default at the start of every turn. Set to true by default.

### Public api:

`Authentication.ts`
```ts
/**
 * User authentication service.
 */
export class Authentication<TState extends TurnState = DefaultTurnState> {
    public readonly settings: OAuthPromptSettings;

    /**
     * Creates a new instance of the `Authentication` class.
     * @param {Application} app - The application instance.
     * @param {string} name - The name of the connection.
     * @param {OAuthPromptSettings} settings - Authentication settings.
     * @param {Storage} storage - A storage instance otherwise Memory Storage is used.
     * @param {MessagingExtensionAuthentication} messagingExtensionsAuth - Handles messaging extension flow authentication.
     * @param {BotAuthentication} botAuth - Handles bot-flow authentication.
     */
    constructor(
        app: Application<TState>,
        name: string,
        settings: OAuthPromptSettings,
        storage?: Storage,
        messagingExtensionsAuth?: MessagingExtensionAuthentication,
        botAuth?: BotAuthentication<TState>
    );

    /**
     * Signs in a user.
     *
     * This method will be called automatically by the Application class.
     * @template TState
     * @param {TurnContext} context - Current turn context.
     * @param {TState} state Application state.
     * @returns {string | undefined} The authentication token or undefined if the user is still login in.
     */
    public signInUser(context: TurnContext, state: TState): Promise<string | undefined>;

    /**
     * Signs out a user.
     * @template TState
     * @param {TurnContext} context - Current turn context.
     * @param {TState} state - Application state.
     * @returns {Promise<void>} A Promise representing the asynchronous operation.
     */
    public signOutUser(context: TurnContext, state: TState): Promise<void>;

    /**
     * Check is the user is signed, if they are then returns the token.
     * @param {TurnContext} context Current turn context.
     * @returns {string | undefined} The token string or undefined if the user is not signed in.
     */
    public isUserSignedIn(context: TurnContext): Promise<string | undefined>;

    /**
     * The handler function is called when the user has successfully signed in.
     * This only applies if sign in was initiated by the user sending a message to the bot.
     * This handler will not be triggered if a messaging extension triggered the authentication flow.
     * @template TState
     * @param {(context: TurnContext, state: TState) => Promise<void>} handler The handler function to call when the user has successfully signed in
     */
    public onUserSignInSuccess(handler: (context: TurnContext, state: TState) => Promise<void>): Promise<void>;

    /**
     * This handler function is called when the user sign in flow fails.
     * This only applies if sign in was initiated by the user sending a message to the bot.
     * This handler will not be triggered if a messaging extension triggered the authentication flow.
     * @template TState
     * @param {(context: TurnContext, state: TState, error: AuthError) => Promise<void>} handler The handler function to call when the user failed to signed in.
     */
    public onUserSignInFailure(
        handler: (context: TurnContext, state: TState, error: AuthError) => Promise<void>
    ): Promise<void>;
}
```

`AuthenticationManager.ts`
``` ts
/**
 * The user authentication manager.
 */
export class AuthenticationManager<TState extends TurnState = DefaultTurnState> {
    /**
     * Creates a new instance of the `AuthenticationManager` class.
     * @param {Application} app - The application instance.
     * @param {AuthenticationOptions} options - Authentication options.
     * @param {Storage} storage - A storage instance otherwise Memory Storage is used.
     */
    constructor(app: Application<TState>, options: AuthenticationOptions, storage?: Storage);

    /**
     * @template TState
     * Gets the authentication instance for the specified connection name.
     * @param {string} name The setting name.
     * @returns {Authentication<TState>} The authentication instance.
     */
    public get(name: string): Authentication<TState>;

    /**
     * Signs in a user.
     * @template TState
     * @param {TurnContext} context The turn context.
     * @param {TState} state The turn state.
     * @param {string} settingName Optional. The name of the setting to use. If not specified, the default setting name is used.
     * @returns {Promise<SignInResponse>} The sign in response.
     */
    public signUserIn(context: TurnContext, state: TState, settingName?: string): Promise<SignInResponse>;

    /**
     * Signs out a user.
     * @template TState
     * @param {TurnContext} context The turn context.
     * @param {TState} state The turn state.
     * @param {string} settingName Optional. The name of the setting to use. If not specified, the default setting name is used.
     */
    public signOutUser(context: TurnContext, state: TState, settingName?: string): Promise<void>;
```

`AuthenticationOptions`
```ts
/**
 * The options to configure the authentication manager
 */
export interface AuthenticationOptions {
    /**
     * The authentication settings.
     * Key uniquely identifies the connection string.
     */
    settings: { [key: string]: OAuthPromptSettings };

    /**
     * Describes the setting the bot should use if the user does not specify a setting name.
     */
    default?: string;

    /**
     * Defaults to true.
     * Indicates whether the bot should start the sign in flow when the user sends a message to the bot or triggers a message extension.
     * If set to false, the bot will not start the sign in flow before routing the activity to the bot logic.
     *
     * To set custom logic, set this property to the selector function.
     */
    autoSignIn?: boolean | Selector;
}
```
 
#### Task list

- [x] Sso-auth bot sample
- [x] Sso-auth message extension sample
- [x] Support for multiple auth configurations (ex. AADv2, github auth...etc)
- [x] Unit tests


js/packages/teams-ai/src/authentication/Authentication.ts
```diff
@@ -0,0 +1,110 @@
...
+     * This handler will not be triggered if a messaging extension triggered the authentication flow.
+     * @template TState
+     * @param {(context: TurnContext, state: TState) => Promise<void>} handler The handler function to call when the user has successfully signed in
+     */
+    public async onUserSignIn(handler: (context: TurnContext, state: TState) => Promise<void>): Promise<void> {
+        this._botAuth.onUserSignIn(handler);
+    }
+
+    private isBotAuthFlow(context: TurnContext): boolean {
+        return context.activity.type == ActivityTypes.Message;
```
blackchoey, line 104: The Adaptive Cards submit event is also a message activity: https://github.com/microsoft/teams-ai/blob/main/js/packages/teams-ai/src/AdaptiveCards.ts#L346
I don't know whether there are more cases, may need to update the constraint here.

js/packages/teams-ai/src/authentication/MessagingExtensionAuthentication.ts
```diff
@@ -0,0 +1,111 @@
...
+        const tokenResponse = await UserTokenAccess.getUserToken(context, this._oauthPromptSettings, magicCode);
+
+        if (!tokenResponse || !tokenResponse.token) {
+            // There is no token, so the user has not signed in yet.
+            // Retrieve the OAuth Sign in Link to use in the MessagingExtensionResult Suggested Actions
+
+            const signInResource = await UserTokenAccess.getSignInResource(context, this._oauthPromptSettings);
+            const signInLink = signInResource.signInLink;
+            // Do 'silentAuth' if this is a composeExtension/query request otherwise do normal `auth` flow.
+            const authType = context.activity.name === QUERY_INVOKE_NAME ? 'silentAuth' : 'auth';
```
blackchoey, line 49: If developer does not enable SSO, using `silentAuth` (which means using SSO) may fail?

js/packages/teams-ai/src/authentication/BotAuthentication.ts
```diff
@@ -0,0 +1,188 @@
...
+
+                await context.sendActivity({
+                    value: { status: 200 } as InvokeResponse,
+                    type: ActivityTypes.InvokeResponse
+                });
+
+                // Successful sign in
+                await this._userSignInHandler?.(context, state);
+            } else {
+                await context.sendActivity({
```
blackchoey, line 142: Seems OAuthPrompt has retry mechanism: https://github.com/microsoft/botbuilder-js/blob/main/libraries/botbuilder-dialogs/src/prompts/oauthPrompt.ts#L281. Will this impact how OAuthPrompt works?

js/packages/teams-ai/src/Application.ts
```diff
@@ -61,6 +63,11 @@ export interface ApplicationOptions<TState extends TurnState> {
      */
     adapter?: BotAdapter;
 
+    /**
+     * Optional. OAuth prompt settings to use for authentication.
+     */
+    authentication?: OAuthPromptSettings;
```
blackchoey, line 69: I’m not sure whether using “OAuthPromptSettings” directly as authentication config will confuse developers, especially when they have some knowledge about BotFramework. For example, they might feel it's for bot authentication only. On the opposite, for a developer that only want to build a Message Extension and don't know BotFramework, the extra settings might also be confusing because they don't work (I feel only `ConnectionName` is valid to Message Extension).

Do you think reorganizing the config object is better? Here's a rough thought in my mind to start the brainstorm:
``` json
{
  "connectionName": "xxx",
  "botAuthenticationSettings": {
    "title": "..."
  },
  "messageExtensionAuthenticationSettings": {
    // Should be no settings for ME, we can remove this field. Put it here to better share my thought
  }
}
```

js/packages/teams-ai/src/authentication/BotAuthentication.ts
blackchoey: By the way, the [authentication sample](https://github.com/microsoft/BotBuilder-Samples/tree/main/samples/javascript_nodejs/18.bot-authentication) I saw previously does not return 400. If it's not necessary return a response, do you think we may remove this? The major purpose is keep things simple.

js/packages/teams-ai/src/authentication/MessagingExtensionAuthentication.ts
blackchoey: Maybe we need a way to determine which flow (SSO / non-SSO) to use?

js/packages/teams-ai/src/authentication/Authentication.ts
singhk97: Good catch. Added check to make sure the `context.activity.text` is populated with a non-empty string.

js/packages/teams-ai/src/authentication/MessagingExtensionAuthentication.ts
singhk97: Great catch. I have a feeling that if SSO isn't enabled Teams client should be able to just default to auth even if 'silentAuth' property is used.

js/packages/teams-ai/src/authentication/BotAuthentication.ts
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    public async authenticate(context: TurnContext, state: TState): Promise<string | undefined> {
+        // Get property names to use
+        const userAuthStatePropertyName = this.getUserAuthStatePropertyName(context);
+        const userDialogStatePropertyName = this.getUserDialogStatePropertyName(context);
+
+        // Save message if not signed in
+        if (!state.conversation.value[userAuthStatePropertyName]) {
+            state.conversation.value[userAuthStatePropertyName] = {
```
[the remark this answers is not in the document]
singhk97: `signedIn` property is removed. Bot sso flow always ends with an incoming `signIn/verifyState` activity, because of that the initial message the user sent needs to be saved and then restored.

js/packages/teams-ai/src/Application.ts
singhk97: I have made some major refactors since this comment. Let me know what you think about the current shape of `AuthenticationOptions`.

js/packages/teams-ai/src/authentication/MessagingExtensionAuthentication.ts
singhk97: if we send `silentAuth` then Teams client trigger token exchange flow. When the bot tries to do token exchange it will fail and elegantly return 412 response. To which Teams client will trigger the auth flow.

js/packages/teams-ai/src/authentication/BotAuthentication.ts
singhk97: The OAuthPrompt's retry mechanism is disabled by default. I throw an "AuthError" and let the application code decide whether to retry or not.

The only exception to this is that in the `botAuth.authenticate()` method if the dialog completes without a token, this means the user had cancelled the SSO flow after the previous turn. In this case we should clear dialog state and retry authentication: https://github.com/microsoft/teams-ai/pull/670/files#diff-5eec9e5c9f4146ed5353d5c5b52d82410317124563e16962d168a7a3d0d104d1R112-R116

singhk97: Good call @blackchoey, because the invoke response should be set from the OAuthPrompt dialog continue through the `runDialog()` call. Removed the invoke responses.

js/packages/teams-ai/src/authentication/Authentication.spec.ts
```diff
@@ -0,0 +1,201 @@
+// import * as sinon from 'sinon';
+// import { BotAuthentication } from './BotAuthentication';
+// import { MessagingExtensionAuthentication } from './MessagingExtensionAuthentication';
+// import { Application } from '../Application';
+// import { OAuthPromptSettings } from 'botbuilder-dialogs';
```
corinagum, line 5: Why is this file commented out?

singhk97: i'm still implementing the test files, you can ignore this for now. The next couple commits will replace this with unit tests.

js/samples/06.auth.botSSO/README.md
```diff
@@ -0,0 +1,156 @@
...
+    ```
+
+1. Copy the ngrok URL and put the URL and domain in the `/env/env.local` file
+
+    ```bash
+    BOT_ENDPOINT=https://{ngrok-url}.ngrok.io
+    BOT_DOMAIN={ngrok-url}.ngrok.io
+    ```
+
+1. In the repository directory, run the Teams Toolkit CLI commands to automate the setup needed for the app
```
corinagum, line 90: Please update readme content to match other samples - we are now redirecting some of the directions to global docs for sanity sake

js/samples/06.auth.messagingExtensionSSO/src/index.ts
```diff
@@ -0,0 +1,233 @@
...
+
+    return {
+        card: profileCard,
+        heigth: 250,
+        width: 400,
+        title: 'Show Profile Card'
+    };
+});
+
+// Matches any link that ends in `.com`.
```
corinagum, line 183: ?

js/samples/06.auth.messagingExtensionSSO/src/index.ts
```diff
@@ -0,0 +1,233 @@
...
+
+    const user = await getUserDetailsFromGraph(token);
+    const profileCard = CardFactory.thumbnailCard(user.displayName, CardFactory.images([user.profilePhoto]));
+
+    return {
+        type: 'result',
+        attachments: [profileCard],
+        attachmentLayout: 'list'
+    };
+});
```
corinagum, line 198: Should type these returns

js/samples/06.auth.messagingExtensionSSO/src/index.ts
```diff
@@ -0,0 +1,233 @@
...
+// Listen for item tap
+app.messageExtensions.selectItem(async (_context: TurnContext, _state: DefaultTurnState, item) => {
+    // Generate detailed result
+    const card = createNpmPackageCard(item);
+
+    // Return results
+    return {
+        attachmentLayout: 'list',
+        attachments: [card],
+        type: 'result'
```
corinagum, line 209: Please see above

js/samples/06.auth.messagingExtensionSSO/src/index.ts
```diff
@@ -0,0 +1,233 @@
...
+
+    const user = await getUserDetailsFromGraph(token);
+    const profileCard = createUserProfileCard(user.displayName, user.profilePhoto);
+
+    return {
+        card: profileCard,
+        heigth: 250,
+        width: 400,
+        title: 'Show Profile Card'
+    };
```
corinagum, line 180: typings

js/samples/06.auth.messagingExtensionSSO/src/index.ts
```diff
@@ -0,0 +1,233 @@
...
+    } else {
+        const response = await axios.get(
+            `http://registry.npmjs.com/-/v1/search?${new URLSearchParams({
+                size: count.toString(),
+                text: searchQuery
+            }).toString()}`
+        );
+
+        // Format search results
+        response?.data?.objects?.forEach((obj: any) => results.push(createNpmSearchResultCard(obj.package)));
```
corinagum, line 122: Should function name be updated?

js/samples/06.auth.messagingExtensionSSO/src/simpleGraphClient.ts
```diff
@@ -0,0 +1,56 @@
...
+    _token: string;
+
+    constructor(token: string) {
+        if (!token || !token.trim()) {
+            throw new Error('SimpleGraphClient: Invalid token received.');
+        }
+
+        this._token = token;
+
+        // Get an Authenticated Microsoft Graph client using the token issued to the user.
```
corinagum, line 22:
```suggestion
        // Get an authenticated Microsoft Graph client using the token issued to the user.
```

js/samples/06.auth.messagingExtensionSSO/src/simpleGraphClient.ts
```diff
@@ -0,0 +1,56 @@
...
+// Licensed under the MIT License.
+
+import { Client } from '@microsoft/microsoft-graph-client';
+import 'isomorphic-fetch';
+
+/**
+ * This class is a wrapper for the Microsoft Graph API.
+ * See: https://developer.microsoft.com/en-us/graph for more information.
+ */
+export class SimpleGraphClient {
```
corinagum, line 11: Is `simple` necessary?

js/samples/06.auth.botSSO/README.md
corinagum: (Same for ME readme)

js/samples/06.auth.botSSO/README.md
```diff
@@ -0,0 +1,156 @@
...
+3. In a terminal, navigate to the sample root.
+
+    ```bash
+    cd teams-ai/js/samples/01.messaging.a.SsoAuthBot/
+    yarn start
+    ```
+
+## Interacting with the bot
+
+You can interact with this bot by sending it a message, which will echo back to you.
```
corinagum, line 46: Fix

js/samples/06.auth.botSSO/README.md
```diff
@@ -0,0 +1,156 @@
+# Teams SSO Bot
+
+This sample shows how to incorporate a basic conversational SSO flow into a Microsoft Teams application using [Bot Framework](https://dev.botframework.com) and the Teams AI SDK.
```
corinagum, line 3: How should the user interact with the bot, and what should they expect? What are the pre-reqs for getting auth to work?

js/samples/06.auth.botSSO/README.md
```diff
@@ -0,0 +1,156 @@
...
+
+
+## Deploy the bot to Azure
+
+You can use Teams Toolkit for VS Code or CLI to host the bot in Azure. The sample includes Bicep templates in the `/infra` directory which are used by the tools to create resources in Azure. Use the **Provision** and **Deploy** menus of the Teams Toolkit extension or from the CLI with `teamsfx provision` and `teamsfx deploy`. [Visit the documentation](https://learn.microsoft.com/en-us/microsoftteams/platform/toolkit/provision) for more info on hosting your app in Azure with Teams Toolkit.
+
+Alternatively, you can learn more about deploying a bot to Azure manually in the [Deploy your bot to Azure](https://aka.ms/azuredeployment) documentation.
+
+## Further reading
+
```
corinagum, line 154: Auth/SSO docs?

js/samples/06.auth.botSSO/README.md
```diff
@@ -0,0 +1,156 @@
...
+    ```
+
+### Manually upload the app to a Teams desktop client
+
+> If you used Teams Toolkit in the above steps, you can [upload a custom app](https://learn.microsoft.com/en-us/microsoftteams/platform/concepts/deploy-and-publish/apps-upload) to a desktop client using the `/appPackage/appPackage.local.zip` file created by the tools and skip to step 6.
+
+1. In a terminal, navigate to `teams-ai/js/samples/01.messaging.a.SsoAuthBot/`
+
+    ```bash
+    cd teams-ai/js/samples/01.messaging.a.SsoAuthBot/
```
corinagum, line 117: ?

js/packages/teams-ai/src/Application.ts
```diff
@@ -148,12 +155,17 @@ export type ConversationUpdateEvents =
...
+ */
+export type Selector = (context: TurnContext) => Promise<boolean>;
+
 /**
  * Function for selecting whether a route handler should be triggered.
  * @param context Context for the current turn of conversation with the user.
  * @returns A promise that resolves with a boolean indicating whether the route handler should be triggered.
  */
-export type RouteSelector = (context: TurnContext) => Promise<boolean>;
+export type RouteSelector = Selector;
```
corinagum, line 168: Why the refactor?

js/packages/teams-ai/src/Application.ts
```diff
@@ -471,6 +527,23 @@ export class Application<TState extends TurnState = DefaultTurnState> {
                 const { storage, turnStateManager } = this._options;
                 const state = await turnStateManager!.loadState(storage, context);
 
+                // Sign the user in
+                if (this._authentication && (await this._startSignIn?.(context))) {
+                    const response = await this._authentication.signUserIn(context, state);
+
+                    if (response.status == 'pending') {
+                        // Save turn state
+                        // - This lets the bot keep track of authentication state
```
corinagum, line 536:
```suggestion
                        // - Save authentication status in turn state 
```

js/packages/teams-ai/src/Application.ts
```diff
@@ -471,6 +527,23 @@ export class Application<TState extends TurnState = DefaultTurnState> {
                 const { storage, turnStateManager } = this._options;
                 const state = await turnStateManager!.loadState(storage, context);
 
+                // Sign the user in
+                if (this._authentication && (await this._startSignIn?.(context))) {
+                    const response = await this._authentication.signUserIn(context, state);
+
+                    if (response.status == 'pending') {
+                        // Save turn state
```
corinagum, line 535: delete

js/packages/teams-ai/src/Application.ts
```diff
@@ -471,6 +527,23 @@ export class Application<TState extends TurnState = DefaultTurnState> {
...
+                        // - This lets the bot keep track of authentication state
+                        await turnStateManager!.saveState(storage, context, state);
+                        return false;
+                    }
+
+                    // Activities for which auth cannot be initiated should be ignored
```
corinagum, line 541: This isn't ignoring though, this is throwing the app.

js/packages/teams-ai/src/Application.ts
```diff
@@ -1047,6 +1132,23 @@ function createMessageReactionSelector(event: MessageReactionEvents): RouteSelec
     }
 }
 
+/**
+ * Returns a selector function that indicates whether the bot should initiate a sign in.
+ * @param {boolean | Selector} startSignIn A boolean or function that indicates whether the bot should initiate a sign in.
+ * @returns {Selector} A selector function that returns true if the bot should initiate a sign in.
+ */
+function createSignInSelector(startSignIn?: boolean | Selector): Selector {
```
corinagum, line 1140: Why is boolean needed? shouldn't it just be the selector?

js/packages/teams-ai/src/MessageExtensions.ts
```diff
@@ -24,34 +24,34 @@ import { Application, RouteSelector, Query } from './Application';
 import { TurnState } from './TurnState';
 
 /**
- * @private
+ * @public
  */
-const ANONYMOUS_QUERY_LINK_INVOKE_NAME = `composeExtension/anonymousQueryLink`;
+export const ANONYMOUS_QUERY_LINK_INVOKE_NAME = `composeExtension/anonymousQueryLink`;
```
corinagum, line 29: Why do these need to be exported?

js/packages/teams-ai/src/authentication/BotAuthentication.ts
```diff
@@ -0,0 +1,280 @@
...
+    verifyStateOperationName,
+    tokenExchangeOperationName
+} from 'botbuilder';
+import { TurnStateProperty } from '../TurnStateProperty';
+import { AuthError } from './Authentication';
+
+/**
+ * @internal
+ */
+interface OAuthPromptResult {
```
blackchoey, line 28: Minor: there's a TokenResponse data model from `botbuilder`, and OAuthPrompt returns it as result. May reuse this data model instead of creating our own data model?

```
import {TokenResponse} from 'botbuilder';
```

js/packages/teams-ai/src/authentication/MessagingExtensionAuthentication.ts
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        const value = context.activity.value;
+
+        // When the Bot Service Auth flow completes, the query.State will contain a magic code used for verification.
+        const magicCode = value.state && Number.isInteger(Number(value.state)) ? value.state : '';
+
+        const tokenResponse = await UserTokenAccess.getUserToken(context, settings, magicCode);
+
+        if (this.isValidActivity(context) && (!tokenResponse || !tokenResponse.token)) {
```
blackchoey, line 39: minor: since the logic will only be called when `Authentication` class found it's a valid Messaging Extension activity, do we still need to do the check here?

js/packages/teams-ai/src/authentication/MessagingExtensionAuthentication.ts
```diff
@@ -0,0 +1,111 @@
...
+            const isTokenExchangable = await this.isTokenExchangeable(context, settings);
+            if (!isTokenExchangable) {
+                await context.sendActivity({
+                    value: { status: 412 } as InvokeResponse,
+                    type: ActivityTypes.InvokeResponse
+                });
+
+                return undefined;
+            } else {
+                const tokenExchangeResponse = await this.exchangeToken(context, settings);
```
blackchoey, line 25: minor: the `isTokenExchangeable(...)` function already did a token exchange, may be able to optimize the code to not do token exchange again.
I'm doing the optimization in my branch, so no action should be required from your side. But if you have any comments regarding the optimization I'm doing, please let me know.

js/packages/teams-ai/src/Application.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable security/detect-object-injection */
```
corinagum, line 1: Remove

js/samples/06.auth.botSSO/README.md
singhk97: There is a separate ticket (#783)  for improving the samples: Added this point to the ticket.

js/samples/06.auth.messagingExtensionSSO/src/simpleGraphClient.ts
singhk97: remove simple

js/samples/06.auth.botSSO/README.md
singhk97: Will be fixed in another ticket: #783

js/samples/06.auth.botSSO/README.md
singhk97: Will be fixed in another ticket: #783

js/samples/06.auth.botSSO/README.md
singhk97: Will be tackled in #784

js/samples/06.auth.botSSO/README.md
singhk97: will be tackled in #783

js/packages/teams-ai/src/Application.ts
singhk97: I'm using `Selector` type to allow users to "select" the activities in which auto sign in should happen. I didn't want to use `RouteSelector` type because it's not selecting a route.

js/packages/teams-ai/src/Application.ts
singhk97: It's ignoring if `response.cause == 'invalidActivity'`

js/packages/teams-ai/src/Application.ts
singhk97: `boolean` is so that users don't have to implement a selector function just to enable/disable it for all incoming activities.

js/packages/teams-ai/src/MessageExtensions.ts
singhk97: It's being used in `MessagingExtensionAuthentication` class to check if incoming invoke activity is one of these types.

js/packages/teams-ai/src/Application.ts
singhk97: this lint issue shows up if i try to do `state.conversation.value[KEY]` where `KEY` is a variable. Apparently this pattern can cause injection attacks. It doesn't apply to our scenario since `KEY` is a constant string.

js/packages/teams-ai/src/Application.ts
corinagum: The code says if it does not equal (`!=`) invalidActivity

js/packages/teams-ai/src/Application.ts
```diff
@@ -471,6 +527,23 @@ export class Application<TState extends TurnState = DefaultTurnState> {
...
+                    const response = await this._authentication.signUserIn(context, state);
+
+                    if (response.status == 'pending') {
+                        // Save turn state
+                        // - Save authentication status in turn state 
+                        await turnStateManager!.saveState(storage, context, state);
+                        return false;
+                    }
+
+                    // Activities for which auth cannot be initiated should be ignored
```
corinagum, line 541:
```suggestion
                    // Invalid activities should be ignored for auth
```

js/packages/teams-ai/src/Application.ts
corinagum: This seems clunky to me, but I don't have simple suggested alternative. Maybe we can bring this up for discussion briefly in a meeting? Doesn't this mean that if they put `false` it will still initiate `startSignIn`? So then isn't it superfluous?

@aacebo do you have thoughts / think this is good as-is?

js/packages/teams-ai/src/Application.ts
corinagum: Do it per line, not per file.

js/packages/teams-ai/src/Application.ts
corinagum: But Selector is the same as what RouteSelector used to be, so how is this different?

js/packages/teams-ai/src/Application.ts
singhk97: yeah so the code says if `response.cause != 'invalidActivity'` then throw an error. This implies that if it's an invalid activity then you ignore. The reason for the comment is that the logic is a bit convoluted. 

I can rewrite it to look like
```ts
if (response.status == 'error') {
    if (response.cause == 'invalidActivity') {
        // do nothing.
    } else {
        throw error;
    }
}
```

I prefer the style I implemented it in because it's concise. Which do you prefer?

js/packages/teams-ai/src/Application.ts
singhk97: There's a bunch of files with the at the top of the file. Can we just tweak the linter to ignore this warning?

js/packages/teams-ai/src/MessageExtensions.ts
corinagum: I copy-pasted these values in MessageExtensions.spec.ts. since there's three places now that they're being used, I'd prefer if we made an enum.

js/packages/teams-ai/src/Application.ts
corinagum: If that's true, I failed to catch it on in-going and they should be removed and added inline as well. If you want to have the rule removed, please set up a meeting with the team to discuss.

js/packages/teams-ai/src/Application.ts
singhk97: The difference is when users hover over the property they'll see `Selector` instead of `RouteSelector` in the tooltip documentation. This is to avoid associating the a generic `Selector` with the specific-purpose `RouteSelector` which represents functions for _selecting_ which _route_ to execute. With a generic `Selector` you're simply _selecting_ certain activities, and the added context comes from where it's used. In this case its used as a type for the `autoSignIn` property in the `AuthenticationOptions` interace.

js/packages/teams-ai/src/Application.ts
corinagum: Okay. I guess the comment + the way the code is done confused me. I do like the conciseness but in situations like this the comment should preclude any ambiguity. How about: 

// Invalid activities will be ignored. Activities that cannot initiate auth will throw an error.

js/packages/teams-ai/src/Application.ts
singhk97: If they put `false` then a selector function will be returned the resolves to false. This means that auth will not be triggered.

Also, this pattern of overloading the parameters is used in virtually every selector creation method:

image

js/packages/teams-ai/src/DefaultTurnStateManager.ts
```diff
@@ -36,6 +36,13 @@ export interface DefaultUserState {}
  * state.
  */
 export interface DefaultTempState {
+    /**
+     * Token returned if the Application was configured with authentication support.
+     */
+    authTokens: {
```
corinagum, line 42: Clarifying question: Does this mean that users will be able to do multiple auth events that need the token in one turn? And then, that means every new turn will need a new auth request/token, correct?

js/packages/teams-ai/src/Application.ts
corinagum: I would say distinguishing between `RouteSelector` and `ActivitySelector` (instead of 'just' `Selector` is valuable here. But if you think it's super obvious and I'm just dumb, then feel free to resolve this thread :)

js/packages/teams-ai/src/Application.ts
singhk97: The two sentences mean two opposite things. How about:

// Invalid activities will be ignored. Any other errors will be re thrown.

js/packages/teams-ai/src/Application.ts
singhk97: Due to time constraints, I'll stick to making it inline. For other places where it's done on top of the file, I'll create a new ticket.

js/packages/teams-ai/src/DefaultTurnStateManager.ts
singhk97: For signing in, yes. Once the user is signed in the bot will simply get the 2+ tokens from the token store with 2 method calls. This will happen in within the same turn.

js/packages/teams-ai/src/TurnStateProperty.ts
```diff
@@ -0,0 +1,63 @@
+/**
+ * @module teams-ai
```
corinagum, line 2: I'm a little concerned this file is superfluous. TurnState.ts already has `deleteValue`, `hasValue`, `getValue`, and `setValue`. 

Is there a difference in functionality because this is using `any` instead of `unknown`?

js/packages/teams-ai/src/authentication/Authentication.ts
```diff
@@ -0,0 +1,292 @@
+/* eslint-disable security/detect-object-injection */
```
corinagum, line 1: Per line instead of per file. For files that already have this, could you file an issue, and we can discuss in a meeting?

js/packages/teams-ai/src/Application.ts
corinagum: Resolving - confusion was that I was magically skipping over line 1144 when trying to parse. Thanks for explaining.

js/packages/teams-ai/src/Application.ts
corinagum: Resolving for now - I will ruminate on our 1:1 call and if I think of a different way to distinguish Selector, we can discuss making an update then :)

js/packages/teams-ai/src/Application.ts
corinagum: Okay to file the issue and keep these files included in that list instead of adding inline for files in this PR.

js/packages/teams-ai/src/TurnStateProperty.ts
corinagum: (Per our 1:1 call) Here could you leave a comment that this is a wrapper specifically for Dialogs, since it needs `StatePropertyAccessor`? TY!

js/packages/teams-ai/src/authentication/BotAuthentication.ts
```diff
@@ -0,0 +1,280 @@
...
+                        context,
+                        state,
+                        new AuthError('Authentication flow completed without a token.', 'completionWithoutToken')
+                    );
+                }
+            }
+        } catch (e) {
+            const errorMessage = e instanceof Error ? e.message : JSON.stringify(e);
+            const message = `Unexpected error encountered while signing in: ${errorMessage}. 
+                Incomming activity details: type: ${context.activity.type}, name: ${context.activity.name}`;
```
corinagum, line 183:
```suggestion
                Incoming activity details: type: ${context.activity.type}, name: ${context.activity.name}`;
```

js/packages/teams-ai/src/authentication/BotAuthentication.ts
```diff
@@ -0,0 +1,280 @@
...
+
+            await this._userSignInFailureHandler?.(context, state, new AuthError(message));
+        }
+    }
+
+    public async runDialog(
```
corinagum, line 189: nit: rename to `runAuthDialog`?

corinagum: feel free to resolve if no

js/packages/teams-ai/src/authentication/BotAuthentication.ts
```diff
@@ -0,0 +1,280 @@
...
+            results = await dialogContext.beginDialog(this._oauthPrompt.id);
+        }
+        return results;
+    }
+
+    public deleteAuthFlowState(context: TurnContext, state: TState) {
+        // Delete user auth state
+        const userAuthStatePropertyName = this.getUserAuthStatePropertyName(context);
+        if (state.conversation.value[userAuthStatePropertyName]) {
+            delete state.conversation.value[userAuthStatePropertyName];
```
corinagum, line 209: Please take note: I think Steve merged in his latest PR that gets rid of the need to add `value` - so you might want to verify before this PR gets checked in.

js/samples/06.auth.botSSO/src/index.ts
```diff
@@ -0,0 +1,135 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
corinagum, line 1: Just FYI - I think the 06 slot got nabbed by Steve for agents. We will need to do a samples refactor anyway, so no need to make a new Auth folder unless you have the extra time.

js/packages/teams-ai/src/MessageExtensions.ts
singhk97: fixed

js/packages/teams-ai/src/authentication/BotAuthentication.ts
singhk97: it's in the BotAuthentication class, so the context should be clear.

js/packages/teams-ai/src/authentication/BotAuthentication.ts
singhk97: Yeah i'll have to merge his changes in.

js/samples/06.auth.botSSO/src/index.ts
singhk97: i'll change it to 07

js/packages/teams-ai/src/TurnStateProperty.ts
singhk97: done

js/packages/teams-ai/src/Application.ts
singhk97: Created issue: [#801](https://github.com/microsoft/teams-ai/issues/801). We don't need to keep a track of the files. We can just do a global CTRL+F to figure out where all it is.

js/packages/teams-ai/src/authentication/BotAuthentication.ts
singhk97: great catch. fixed.

js/packages/teams-ai/src/authentication/MessagingExtensionAuthentication.ts
singhk97: Yeah, for instance let's say the incoming activity is `composeExtension/submitAction`. If there's a token in token store, then this method will return the token. But if it doesn't then it will try to trigger the sign in flow, which is undesirable. And one could argue, makes the `MessagingExtension.authenticate` method buggy.

You can argue that we check if the user is logged in in the `Authenticate.signUserIn` method which is redundant and can be removed. But I put it in for readability and clarity sakes.

js/packages/teams-ai/src/authentication/MessagingExtensionAuthentication.ts
singhk97: yes please refactor this if it makes sense.

corinagum (approved):

PR #814, opened 2023-11-15
[JS] feat: TeamsBotSsoPrompt based SSO support

blackchoey:
## Linked issues

closes: #781 #782

## Details

1. Refactored current authentication design to accept different implementation.
    * Extracted common logic to a base class
    * Messaging extension authentication method no longer accepts a settings parameter, it's already provided in constructor
    * Split `runDialog` to `runDialog` and `continueDialog`, to better handle the events
    * Minor update to logic around Messaging extension magic code to make the flow more clear
2. Add TeamsBotSsoPrompt based SSO implementation.
3. Add test cases

TODO:
1. Add samples (will create another PR)

## Attestation Checklist

  - Local testing
  - E2E testing in Teams


js/packages/teams-ai/src/authentication/Authentication.ts
```diff
@@ -127,6 +160,53 @@ export class Authentication<TState extends TurnState> {
     public onUserSignInFailure(handler: (context: TurnContext, state: TState, error: AuthError) => Promise<void>) {
         this._botAuth.onUserSignInFailure(handler);
     }
+
+    private isOAuthPromptSettings(settings: OAuthPromptSettings | TeamsSsoPromptSettings): settings is OAuthPromptSettings {
+        return (settings as OAuthPromptSettings).connectionName !== undefined;
+    }
+
+    private async getLoginHint(context: TurnContext): Promise<string | undefined> {
```
corinagum, line 168: Here, the return type can be promise of undefined, but on line 178, it can be null. Is there a reason for the difference?

js/packages/teams-ai/src/authentication/MessagingExtensionAuthenticationBase.ts
```diff
@@ -0,0 +1,95 @@
+import { ActivityTypes, InvokeResponse, TokenResponse, TurnContext } from 'botbuilder';
+import { MessagingExtensionsInvokeNames } from '../MessageExtensions';
+
+/**
+ * @internal
+ */
+export abstract class MessagingExtensionAuthenticationBase {
```
corinagum, line 7: Isn't the naming `MessageExtension`? Or is it different for authentication?

corinagum: [Documentation](https://learn.microsoft.com/en-us/microsoftteams/platform/messaging-extensions/how-to/add-authentication) would indicate the correct term is `MessageExtension`:
image

Could we rename away from `MessagingExtension`?

corinagum: Uhhh nevermind. Botbuilder has imports that are also `MessagingExtension` instead of `MessageExtension` so there's no way to fully fix the inconsistency.

js/packages/teams-ai/src/authentication/OAuthPromptBotAuthentication.ts
```diff
@@ -0,0 +1,72 @@
...
+import { Application } from "../Application";
+import { TurnState } from "../TurnState";
+import { TurnStateProperty } from "../TurnStateProperty";
+
+export class OAuthPromptBotAuthentication<TState extends TurnState> extends BotAuthenticationBase<TState> {
+    private _oauthPrompt: OAuthPrompt;
+
+    public constructor(
+        app: Application<TState>,
+        oauthPromptSettings: OAuthPromptSettings, // Child classes will have different types for this
```
corinagum, line 16:
```suggestion
        oAuthPromptSettings: OAuthPromptSettings, // Child classes will have different types for this
```
?

js/packages/teams-ai/src/authentication/Authentication.ts
blackchoey: In my mind, a function should return null because it clearly tells something does not have a value. And some library (like MSAL) also follows this rule. So I will try to return null if possible, like what I'm doing on line 178.
But I met an external function that declares loginHint parameter as optional (xxxFunction(loginHint?: string), null is not accepted in this case. Since `getLoginHint` is just a private function, I returned undefined to reduce the coding complexity.

I think this is a coding style thing. If there's some style I need to follow, I will update the PR to satisfy the style.

js/packages/teams-ai/src/authentication/MessagingExtensionAuthenticationBase.ts
```diff
@@ -0,0 +1,95 @@
...
+    public async authenticate(context: TurnContext): Promise<string | undefined> {
```
corinagum, line 8: Sorry for the hassle, but please rename this file and **new** methods/variables/etc with MessageExtension instead of MessagingExtension. We will try to stay consistent about this from now on.

js/packages/teams-ai/src/authentication/OAuthPromptMessagingExtensionAuthentication.ts
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { TokenResponse, TurnContext } from "botbuilder";
+import { OAuthPromptSettings } from "botbuilder-dialogs";
+import { MessagingExtensionAuthenticationBase } from "./MessagingExtensionAuthenticationBase";
+import * as UserTokenAccess from './UserTokenAccess';
+
+export class OAuthPromptMessagingExtensionAuthentication extends MessagingExtensionAuthenticationBase {
```
corinagum, line 9: Please rename to `OAuthPromptMessageExtensionAuthentication`

js/packages/teams-ai/src/authentication/Authentication.ts
corinagum: thanks for the explanation - this looks good as-is to me :)

js/packages/teams-ai/src/authentication/TeamsSsoMessagingExtensionAuthentication.ts
```diff
@@ -0,0 +1,67 @@
...
+
+import { TurnContext } from "botbuilder-core";
+import { TeamsChannelAccount, TokenResponse } from "botframework-schema";
+import { MessagingExtensionAuthenticationBase } from "./MessagingExtensionAuthenticationBase";
+import { TeamsSsoPromptSettings } from "./TeamsBotSsoPrompt";
+import { ConfidentialClientApplication } from "@azure/msal-node";
+import { TeamsInfo } from "botbuilder";
+import { MessagingExtensionsInvokeNames } from '../MessageExtensions';
+
+export class TeamsSsoMessagingExtensionAuthentication extends MessagingExtensionAuthenticationBase {
```
corinagum, line 12: Please rename to `TeamsSsoMessageExtensionAuthentication` (same with any non-external package instances that say MessagingExtension`

corinagum (dismissed):
LGTM! Sorry for the hassle but if you could please do the rename to `...MessageExtension...`, this is the correct terminology and we need to stay as consistent about it as possible, even though some instances must stay as-is.

If you have the permissions to merge after the change, feel free to do so. Otherwise, feel free to ping me any time and I will merge in asap.

singhk97 (approved):

js/packages/teams-ai/src/TurnState.ts
```diff
@@ -57,6 +57,11 @@ export interface DefaultTempState {
      * User authentication tokens
      */
     authTokens: { [key: string]: string };
+
+    /**
+     * Flag indicating whether a token exchange event has already been processed
+     */
+    duplicateTokenExchange?: boolean
```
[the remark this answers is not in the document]
blackchoey: This is used by our library internally. Our library will save the state after processing every request. This will cause ETag conflict when processing the duplicate token exchange activities. So I introduced this flag to let our library be able to skip state saving for duplicated token exchange activity.

I think it can also be adopted by oAuthPrompt. But since we're using the token exchange middleware from botbuilder to handle the duplicate for oAuthPrompt, it's currently only used for TeamsSsoPrompt. I choose the current implementation to reduce the token exchange requests.

corinagum (dismissed):

js/packages/teams-ai/src/TurnState.ts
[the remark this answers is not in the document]
singhk97: Since it's used internally in the library we should ideally obscure this from the end user, just like how state the Bot Authentication state is stored. Does this method scale to multiple concurrent login requests from different users?

singhk97 (dismissed):

singhk97:
I think that `OAuthPrompt` prefix to `OAuthPromptMessageExtensionAuthentication` makes the class name to long. I think that "Prompt" in the name is superfluous because it does inform much other than we're using "OAuthPrompt" dialog under the hood. 

I propose we change "OAuthPrompt" to just "OAuth". And TeamsSSOPrompt to "TeamsSSO" to keep things simple.

singhk97:
Also, let's pick a consistent variable/property naming for "OAuth". Let's stick with `oauth` or `oAuth` but not both. @blackchoey it's your call

lilyydu (approved):

PR #896, opened 2023-11-27
[JS] feat: Who bot sample, add user sign in method for end user

singhk97:
#### work in progress
- [x] Update the sample to include more graph calls and test it

## Linked issues

closes: #890 #895

#### Change details

- Added Who bot sample
- Added `getTokenOrSignInUser` method in `Application` class. It is meant to be the single catch-all method to sign a user in and get the token. If the user is signed in it simply returns the token. If not, it triggers the sign in flow. Starting the sign in flow will only work if the incoming activity supports it (ex. `message`, `messageExtension/fetch`...etc). This method will work in action handlers as well.

## Attestation Checklist

  - Local testing
  - E2E testing in Teams


js/packages/teams-ai/src/Application.ts
```diff
@@ -938,6 +962,59 @@ export class Application<TState extends TurnState = TurnState> {
...
+     * @summary
+     * Use this method to get the access token for a user that is signed in to the bot.
+     * If the user isn't signed in, this method starts the sign-in flow.
+     * The bot should end the turn in this case until the sign-in flow completes and the user is signed in.
+     * @param {TurnContext} context - The context for the current turn with the user.
+     * @param {TState} state - The current state of the conversation.
+     * @param {string} settingName The name of the authentication setting.
+     * @returns {string | undefined} The token for the user if they are signed in, otherwise undefined.
+     */
+    public async getTokenOrStartSignIn(
```
singhk97, line 978: @blackchoey this is meant to be the method the end-user calls if they want to trigger sign in. It will put the user in "sign in" state and the subsequent incomming activities will be routed to the authentication logic in the `Application.run()` until sign in flow completes - either with a success or failure.

js/samples/07.whoBot/src/bot.ts
```diff
@@ -0,0 +1,269 @@
...
+if (!process.env.OPENAI_KEY && !process.env.AZURE_OPENAI_KEY) {
+    throw new Error('Missing environment variables - please check that OPENAI_KEY or AZURE_OPENAI_KEY is set.');
+}
+
+export const run = (context: TurnContext) => app.run(context);
+
+// Create AI components
+const model = new OpenAIModel({
+    // OpenAI Support
+    apiKey: process.env.OPENAI_KEY!,
```
blackchoey, line 26: Should we comment the OpenAI settings? It seems developers only need to config one of OpenAI/AzureOpenAI, configuring both might be confusing. We can keep the commentted OpenAI settings as a reference for developers.

js/samples/07.whoBot/src/bot.ts
```diff
@@ -0,0 +1,269 @@
...
+
+// Register AI actions
+app.ai.action('user_personal_information', async (context: TurnContext, state: ApplicationTurnState) => {
+    const token = await app.getTokenOrStartSignIn(context, state, 'graph');
+    if (!token) {
+        await context.sendActivity('You have to be signed in to fulfill this request. Starting sign in flow...');
+        return AI.StopCommandName;
+    }
+
+    const graphClient = new GraphClient(state.temp.authTokens['graph']!);
```
blackchoey, line 91: Line 85 already gets the token, should we use the `token` variable?

blackchoey:
FYI: not sure whether we should merge the changes to `main` branch

blackchoey (dismissed):
Overall looks good to me. Left a few minor comments.

js/samples/07.whoBot/src/bot.ts
singhk97: We can keep both the settings so that the user can simply decide which one to use by setting the config value in the .env file. If they set `OPENAI_KEY` then open ai service will used. If they set `AZURE_OPENAI_KEY` and `AZURE_OPENAI_ENDPOINT` then azure open ai service will be used.

blackchoey (approved):
Approving first to avoid blocking the PR. But I see there're a lot of other files changed. Seems there should be no C# changes in this PR? Please check with @corinagum before merge to avoid bringing unexpected changes.

singhk97:
The branch was based on `breaking-changes-2` which it shouldn't have. Creating another branch that's based on `main` instead. PR: https://github.com/microsoft/teams-ai/pull/911

Issue #454, opened 2023-08-21
Telemetry Headers


corinagum:
We need to add telemetry for tracking AI Library usage:

Users actively building using AI library - Test vs production
What are bot feature APIs they are using
MAU of app built via SDK

[code](https://github.com/microsoft/teams-ai/blob/stevenic/unification/js/packages/teams-connector/src/auth/connectorFactoryImpl.ts#L12)

Make sure SMBA is logging correctly
```[tasklist]
### Tasks
- [ ] https://github.com/microsoft/teams-ai/issues/598
- [ ] https://github.com/microsoft/teams-ai/issues/745
```

corinagum:
Requests made to botbuilder team:

https://github.com/microsoft/botbuilder-js/issues/4552
https://github.com/microsoft/botbuilder-dotnet/issues/6701
https://github.com/microsoft/botbuilder-python/issues/2033

aacebo:
currently following up with SMBA on our side to make sure they can ingest custom headers we set. BF team should be able to start this work in the JS SDK early next year.

PR #1142, opened 2024-01-09
[JS] feat: add user agent override and internalize `BotAdapter`

aacebo:
## Linked issues

closes: #454 

## Details

In order to configure the `User-Agent` header for telemetry, I have added the class `TeamsBotFrameworkAuthentication` which will allow us to control the headers while still giving the devs implementation flexibility:

- we can control how the bot adapter is initialized and provide overrides when necessary.
- we can add an `error` handler on the `Application` class.
- we can cut down on botbuilder boilerplate necessary for users to use our library.


js/packages/teams-ai/src/Application.ts
```diff
@@ -366,6 +403,19 @@ export class Application<TState extends TurnState = TurnState> {
         return this._taskModules;
     }
 
+    /**
+     * Sets the bots error handler
```
corinagum, line 407:
```suggestion
     * Sets the bot's error handler
```

js/packages/teams-ai/src/Application.ts
```diff
@@ -87,6 +104,17 @@ export interface ApplicationOptions<TState extends TurnState> {
      */
     botAppId?: string;
 
+    /**
+     * Optional. Bot authentication configuration
+     */
+    botAuthentication?: {
```
lilyydu, line 110: May need to add setter for this for `ApplicationBuilder`

js/packages/teams-ai/src/Application.ts
```diff
@@ -87,6 +104,17 @@ export interface ApplicationOptions<TState extends TurnState> {
      */
     botAppId?: string;
 
+    /**
+     * Optional. Bot authentication configuration
```
corinagum, line 108:
```suggestion
     * Optional. Bot authentication configuration for initializing the bot CloudAdapter
```

js/packages/teams-ai/src/Application.ts
```diff
@@ -87,6 +104,17 @@ export interface ApplicationOptions<TState extends TurnState> {
...
+     */
+    botAuthentication?: {
+        botFrameworkAuthConfig?: ConfigurationBotFrameworkAuthenticationOptions;
+        credentialsFactory?: ServiceClientCredentialsFactory;
+        authConfiguration?: AuthenticationConfiguration;
+        botFrameworkClientFetch?: (input: RequestInfo, init?: RequestInit) => Promise<Response>;
+        connectorClientOptions?: ConnectorClientOptions;
```
corinagum, line 115: Might be good to make this an interface?

js/packages/teams-ai/src/Application.ts
```diff
@@ -68,6 +83,8 @@ export interface Query<TParams extends Record<string, any>> {
 export interface ApplicationOptions<TState extends TurnState> {
     /**
      * Optional. Bot adapter being used.
+     * @deprecated
```
corinagum, line 86: We probably want to throw an error or at least set a warning when adapter is being set. The test files set adapter, and it's not a big deal to update but will need to be done, and the reminder/warning will help with that.

corinagum: Would be good to also add migration notes in getting-started since it's a breaking change technically.

js/packages/teams-ai/package.json
```diff
@@ -2,7 +2,7 @@
     "name": "@microsoft/teams-ai",
     "author": "Microsoft Corp.",
     "description": "SDK focused on building AI based applications for Microsoft Teams.",
-    "version": "1.0.1",
+    "version": "1.1.0",
```
singhk97, line 5: do we want to bump the version right before we release it?

aacebo: I have to do it in this PR because it fails the samples CI otherwise (due to sym linking)

aacebo: I did it in this commit because the build was failing https://github.com/microsoft/teams-ai/pull/1142/commits/a338c3dad7859dbec0ed1b70ec7130b34188af6b

js/packages/teams-ai/src/Application.ts
```diff
@@ -67,26 +70,15 @@ export interface Query<TParams extends Record<string, any>> {
...
 export interface ApplicationOptions<TState extends TurnState> {
     /**
-     * Optional. Bot adapter being used.
-     * @remarks
-     * If using the `longRunningMessages` option or calling the continueConversationAsync() method,
-     * this property is required.
+     * Optional. Options used to initialize your `BotAdapter`
      */
-    adapter?: BotAdapter;
+    adapter?: BotAdapterOptions;
```
singhk97, line 75: The authentication samples will break with this change as it uses the adapter.

aacebo: ya, if we go with this implementation I will open another PR to update the samples once we publish the new version. Do we normally update the samples before we publish?

corinagum: It's better to do a fast-follow for samples to decrease the size of PRs. Also gives a little more time for testing.

js/packages/teams-ai/src/MessageExtensions.spec.ts
```diff
@@ -27,7 +28,8 @@ describe('MessageExtensions', () => {
     const adapter = new TestAdapter();
     let mockApp: Application;
     beforeEach(() => {
-        mockApp = new Application({ adapter });
+        mockApp = new Application();
+        sinon.stub(mockApp, 'adapter').get(() => adapter);
```
corinagum, line 32: This actually needs a sandbox (Learned this an hour ago)

```
    let sinonSandbox: sinon.SinonSandbox;
    let createStub: sinon.SinonStub;

// ----

beforeEach(() => {
        sinonSandbox = createSandbox();
        createStub = sinonSandbox.stub(axios, 'create').returns(mockAxios);
});
// ...
    afterEach(() => {
        sinonSandbox.restore();
    });

// test: 
 assert.equal(createStub.called, true);
```

I think there are a number of places this needs to be fixed, not just here, so if you would prefer to file an issue and assign it to me I can make the fixes all at once. :)

corinagum (changes requested):
LGTM! 

Just a couple comments, will approve immediately after. Thanks Alex!

corinagum (approved):

lilyydu (approved):

Issue #1070, opened 2023-12-18
[PY] Embeddings


aacebo:
implement `embeddings` functionality of `JS` SDK https://github.com/microsoft/teams-ai/tree/main/js/packages/teams-ai/src/embeddings

PR #1217, opened 2024-01-30
[PY] feat: embeddings

lilyydu:
## Linked issues

closes: #1070 

## Details

Implemented the logic and tests for embeddings for Python, based on JS.

1.  Client API calls and types are using the [OpenAI Python API library](https://github.com/openai/openai-python/tree/main) 
2. I restructured the interfaces and classes for AzureOpenAI and OpenAI into its own classes. Two reasons- (a) some logic is different, and (b) this allows for better maintainability + response to bugs since they depend on different clients and configurations from the API library.
3. To replicate `setTimeout` in retries, I've used `asyncio.sleep` but open to better preferences/suggestions if any
4. Logging follows C# implementation (no colorization)
5. Had to disable 2 _pylint_ and _mypy_ type/argument checker (`unused-argument` because mock.patch requires the same method signature, and `arg-type` because openai.types doesn't yet export the Usage class- [filed a request here](https://github.com/openai/openai-python/issues/1135))

## Attestation Checklist

  - Local testing
  - E2E testing in Teams


lilyydu:
Closing for #1223 - will open and merge later

python/packages/ai/pyproject.toml
```diff
@@ -26,6 +26,8 @@ pytest-asyncio = "^0.21.1"
 black = "^23.7.0"
 isort = "^5.12.0"
 mypy = "^1.5.0"
+httpx = "^0.26.0"
+openai = "^1.11.1"
```
aacebo, line 30: is openai supposed to be a dev dependency? Seems like we will need it in the models as a real dep. Also we can find a way to not import httpx since we already have aiohttp as an http client installed.

aacebo (approved):

Issue #1065, opened 2023-12-18
[PY] Prompts


aacebo:
implement `prompts` functionality of the `JS` SDK https://github.com/microsoft/teams-ai/tree/main/js/packages/teams-ai/src/prompts

Remaining classes:
- [ ] Prompt
- [ ] PromptManager
- [ ] PromptTemplate
- [ ] UserInputMessage

PR #1233, opened 2024-02-02
feat: add prompt manager

blackchoey:
## Linked issues

closes: #1065

## Details

1. Add `PromptManager` and corresponding classes
2. Accessing `DataSource.name` will throw error with `TextDataSource` instance. Update `TextDataSource` and `DataSource` class to expose the `name` property to fix this issue.
3. Add `from_dict` method to some dataclasses to allow them initialize from a json file

Note: augmentations and actions related logic are not included in this PR as they're not ready yet.

## Attestation Checklist

  - Local testing
  - E2E testing in Teams

### DIFF
--- a/python/packages/ai/teams/ai/data_sources/data_source.py
+++ b/python/packages/ai/teams/ai/data_sources/data_source.py
@@ -17,8 +17,10 @@ class DataSource(ABC):
     A data source that can be used to render text that's added to a prompt.
     """
 
-    name: str
-    "Name of the data source."
+    @property
+    @abstractmethod
+    def name(self) -> str:
+        "Name of the data source."
 
     @abstractmethod
     async def render_data(
--- a/python/packages/ai/teams/ai/data_sources/text_data_source.py
+++ b/python/packages/ai/teams/ai/data_sources/text_data_source.py
@@ -38,7 +38,7 @@ class TextDataSource(DataSource):
         self._text = text
 
     @property
-    def get_name(self) -> str:
+    def name(self) -> str:
         """
         Name of the data source.
         """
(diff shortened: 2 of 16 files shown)


python/packages/ai/teams/ai/data_sources/data_source.py
```diff
@@ -17,8 +17,10 @@ class DataSource(ABC):
     A data source that can be used to render text that's added to a prompt.
     """
 
-    name: str
-    "Name of the data source."
+    @property
+    @abstractmethod
```
lilyydu, line 21: Is `@abstractmethod` equivalent to `return self._name`?

blackchoey: Marking it as `@abstractmethod` lets people determine their own logic to provide the name. For example, they may want to use another variable to store the name, or calculate the name instead of just returning a private property.
Removing `@abstractmethod` and let it `return self._name` you proposed also works, we need to have a constructor in the abstract class to ensure the property is initialized in this case.

Let me know which way you prefer.

aacebo: I think this makes sense, @lilyydu this is essentially in typescript a getter with a default implementation that can be overridden by a class that extends it.

```typescript
export class DataSource {
    get name(): string {
        return "name";
    }
}

export class MyDataSource extends DataSource {
    get name(): string { return this._name; }
    private _name: string;
}
```

lilyydu: Gotcha, this works fine by me- thank you both for explaining!

aacebo (approved):

Issue #1274, opened 2024-02-08
[C#] Port missing properties in ActionPlanner


singhk97:
The [`ActionPlanner.cs`](https://github.com/microsoft/teams-ai/blob/main/dotnet/packages/Microsoft.TeamsAI/Microsoft.TeamsAI/AI/Planners/ActionPlanner.cs) class is missing the following public properties: `defaultModel`, `prompts`, & `model` when compared to it's JS equivalent:

 https://github.com/microsoft/teams-ai/blob/09ebba139104816ea675e6e7958e479c7c3de6d2/js/packages/teams-ai/src/planners/ActionPlanner.ts#L127-L136

TODO:
- [x] Add these properties to the  [`ActionPlanner.cs`](https://github.com/microsoft/teams-ai/blob/main/dotnet/packages/Microsoft.TeamsAI/Microsoft.TeamsAI/AI/Planners/ActionPlanner.cs) class with only a getter. 
- [x] Add unit tests in the [`ActionPlannerTests.cs`](https://github.com/microsoft/teams-ai/blob/main/dotnet/packages/Microsoft.TeamsAI/Microsoft.TeamsAI.Tests/AITests/ActionPlannerTests.cs) class

PR #1303, opened 2024-02-18
[C#] port: missing properties on ActionPlanner

nkrama-99:
## Linked issues

closes: #1274 

## Details
- port missing relevant public gets

### DIFF
--- a/dotnet/packages/Microsoft.TeamsAI/Microsoft.TeamsAI.Tests/AITests/ActionPlannerTests.cs
+++ b/dotnet/packages/Microsoft.TeamsAI/Microsoft.TeamsAI.Tests/AITests/ActionPlannerTests.cs
@@ -474,6 +474,54 @@ namespace Microsoft.Teams.AI.Tests.AITests
             Assert.Equal(planMock.Object, result);
         }
 
+        [Fact]
+        public void Test_Get_Model()
+        {
+            // Arrange
+            var modelMock = new Mock<IPromptCompletionModel>();
+            var prompts = new PromptManager();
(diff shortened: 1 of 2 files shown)


nkrama-99:

singhk97:
@nkrama-99  can you also add a public getting from `defaultPrompt`: https://github.com/microsoft/teams-ai/blob/09ebba139104816ea675e6e7958e479c7c3de6d2/js/packages/teams-ai/src/planners/ActionPlanner.ts#L127-L136

nkrama-99:
@singhk97 defaultPrompt in js project is a "string?" which is only populated if this._options.defaultPrompt is also a string (the property defaultPrompt in ActionPlannerOptions is a "string | ActionPlannerPromptFactory"). In C# project, it can only be of one type, and it's ActionPlannerPromptFactory. If I were to add a public getter for defaultOptions, would it be of type ActionPlannerPromptFactory?

singhk97:
The JS provides more flexibility (string or fn) in setting the default prompt since it supports union types. C# only supports passing in a delegate, which is restrictive but doesn't change the possible values that can be assigned the `DefaultPrompt`.

So yeah, the C# `DefaultPrompt` should be the `ActionPlannerPromptFactory` type. Thanks for clarifying @nkrama-99!

nkrama-99:
@singhk97 thanks for the clarification! I've made the changes.

corinagum (approved):

singhk97 (approved):

singhk97:
Thanks @nkrama-99!

PR #1310, opened 2024-02-20
[PY] feat: deploy echo bot to azure

blackchoey:
## Linked issues

closes: #1254 

## Details

1. Add configuration to allow users deploy the sample to Azure
2. Use `Flask` instead of `FastAPI`, which is easier to be hosted using Azure App Service
3. Use `requirements.txt` instead of `pyproject.toml` to manage dependencies, as Azure App Service does not support pyproject.toml
4. Since the sample uses `requirements.txt` now, remove poetry based CI commands and call corresponding tools in GitHub actions directly. This aims to avoid having both `requirements.txt` and `pyproject.toml`, which may cause confusions. Still work-in-progress, requires updating other samples as well.
5. Remove unused `Attach to Local Service` config in `launch.json`

## Attestation Checklist

- [X] My code follows the style guidelines of this project

  - Local testing
  - E2E testing in Teams


aacebo (changes requested):
I think its best to keep using poetry for all the benefits it gives us as well as consistency, poetry even has a command for generating `requirements.txt` for situations like this, it is
```bash
poetry export --without-hashes --format=requirements.txt > requirements.txt
```

aacebo (approved):

PR #1362, opened 2024-03-01
[PY] feat: ChatModeration Sample

lilyydu:
## Linked issues

closes: #318 

## Details

- ported over Moderation bot from JS 
- updated formatting of config.py, READMEs, yml and ENV setup for all samples
- fixed bugs in AzureContentSafetyModerator
- fixed configuration bugs in AI-ME bot


python/samples/04.ai.b.messageExtensions.AI-ME/env/.env.dev.user
```diff
@@ -1,8 +1,3 @@
 # This file includes environment variables that will not be committed to git by default. You can set these environment variables in your CI/CD system for your project.
 
-# Secrets. Keys prefixed with `SECRET_` will be masked in Teams Toolkit logs.
```
blackchoey, line 3: Can you share some background about removing the content? The purpose to having these content is to let users know what will be generated during provision / local debug, as well as letting them easily fill an existing value if they want.

The Azure OpenAI envs is also required when provision an Azure environment.

python/samples/04.ai.b.messageExtensions.AI-ME/env/.env.dev.user
```diff
@@ -1,8 +1,3 @@
...
-SECRET_BOT_PASSWORD=
-
-# Provide Azure OpenAI API Key and Endpoint
-AZURE_OPENAI_KEY=
-AZURE_OPENAI_ENDPOINT=
\ No newline at end of file
+TEAMS_APP_UPDATE_TIME=
```
blackchoey, line 3: It's an internal state used to optimize the local debug / provision performance, and not expected to be modified by users. So not necessarily to be included in this file. I personally recommend appending `TEAMS_APP_UPDATE_TIME` to the end of existing content if you would like to include it.

python/samples/04.ai.b.messageExtensions.AI-ME/env/.env.local
```diff
@@ -6,6 +6,8 @@ APP_NAME_SUFFIX=local
 
 # Generated during provision, you can also add your own variables.
 BOT_ID=
+SECRET_BOT_PASSWORD=
```
blackchoey, line 9: Suggest moving the secrets to `.user` file, so everything in `.env.local` can be shared without security concern (though it's gitignored by default).

python/samples/04.ai.b.messageExtensions.AI-ME/env/.env.local.user
```diff
@@ -1,4 +1,4 @@
 # This file includes environment variables that will not be committed to git by default. You can set these environment variables in your CI/CD system for your project.
 
-# Secrets. Keys prefixed with `SECRET_` will be masked in Teams Toolkit logs.
```
blackchoey, line 3: Would be better to keep the comments so users can quickly understand the `SECRET_` prefix.

python/samples/04.ai.b.messageExtensions.AI-ME/teamsapp.yml
```diff
@@ -98,4 +98,3 @@ deploy:
       # You can replace it with your existing Azure Resource id
       # or add it to your environment variable file.
       resourceId: ${{BOT_AZURE_APP_SERVICE_RESOURCE_ID}}
-projectId: dde6b9de-663f-4cd4-8c24-164beafa5501
```
blackchoey, line 101: Good catch, thank you!

python/samples/04.ai.b.messageExtensions.AI-ME/env/.env.dev.user
lilyydu: I was thinking we can include this comment in the READMEs instead, but open to keeping it here as well. 

AOAI isn't supported right now (moderation logic requires fixes in the `openai-python` library) so I didn't include it here so as to avoid confusion

python/samples/04.ai.b.messageExtensions.AI-ME/env/.env.dev.user
lilyydu: Where are you referring to by "end of existing content"? I noticed it was autogenerated on my end here so I kept it inside

python/samples/04.ai.b.messageExtensions.AI-ME/env/.env.local
lilyydu: The F5 flow was complaining if I didn't insert it in this file (rather than the `.local.user` file) - do you know where this is being mandated?

aacebo (dismissed):

python/samples/04.ai.b.messageExtensions.AI-ME/env/.env.dev.user
blackchoey: The change looks good to me now. Thank you.

python/samples/04.ai.b.messageExtensions.AI-ME/env/.env.dev.user
blackchoey: Thank you. Let keep the comments in .env files first to align with TTK templates and we can discuss further if needed.

The AOAI/OAI envs in `.env.dev.user` are used to configure Azure App Service app settings. TTK does not consume information from `.env` file in this step as it may contain envs that not related to the `dev` environment. So we still need to add a placeholder to `.env.dev.user` for users to easily provide AOAI/OAI envs for the dev environment.

For AOAI/OAI envs used for local debugging, I think it's already aligned with JS/C# samples by haveing a `sample.env` in the root.

aacebo (dismissed):

aacebo (approved):

PR #1364, opened 2024-03-03
[PY] fix: Turn State

aacebo:
## Linked issues

#minor

## Details

- reimplement turn state to satisfy requirements while supporting customization.
- ensure state if fully serializable for easy of use
- ensure state supports new state properties other than "conversation, user, temp" via `@state` decorator

### Custom State

image

### Turn State Factory

image

### Serialization

image
image


python/samples/04.ai.b.messageExtensions.AI-ME/pyproject.toml
```diff
@@ -11,7 +11,7 @@ packages = [
 [tool.poetry.dependencies]
 python = ">=3.8,<4.0"
 python-dotenv = "^1.0.1"
-teams-ai = "^1.0.0rc1"
+teams-ai = { path = "../../packages/ai", develop = true }
```
blackchoey, line 14: Will we change the reference to a published version in PYPI later?
Deploying to app service does not work with local package because app service needs to pip install the packages from server side.

aacebo: yes we will change it later, right now for the new state changes to work it needs to reference the local version

lilyydu (approved):

Issue #1326, opened 2024-02-22
[PY] Assistants Planner


aacebo:
implement `AssistantsPlanner`  https://github.com/microsoft/teams-ai/blob/main/js/packages/teams-ai/src/planners/AssistantsPlanner.ts

Issue #1252, opened 2024-02-06
[PY] Math Bot


lilyydu:

lilyydu:
completed via https://github.com/microsoft/teams-ai/commit/a25260adee39eba9c474e0cd483f7ff94b3d29c6

PR #1453, opened 2024-03-25
[PY] feat: OAI AssistantsPlanner and MathBot 

lilyydu:
## Linked issues

closes: #1326 and #1252 

## Details

- With reference to JS, implemented support for AssistantsPlanner with OAI and MathBot sample
- updated `Usage` class typing for tests
- will be adding tests in another PR (plan to add it for all our Planners at once)

Compared to JS:
- returning an empty Plan in `submit_user_input` and `submit_action_results`
- removed smaller methods that only had single API call


python/packages/ai/teams/ai/planners/assistants_planner.py
```diff
@@ -0,0 +1,379 @@
...
+        if run.status == "failed":
+            return True
+        if run.status == "cancelled":
+            return True
+        if run.status == "expired":
+            return True
+        return False
+
+    @staticmethod
+    async def create_assistant(
```
aacebo, line 307: can we move this public method up with the other publics? Usually easier to read if publics are above privates

python/packages/ai/teams/ai/planners/planner.py
```diff
@@ -6,14 +6,17 @@
...
 from botbuilder.core import TurnContext
 
 from ...state import TurnState
 from .plan import Plan
 
+StateT = TypeVar("StateT", bound=TurnState)
 
-class Planner(ABC):
+
+class Planner(Generic[StateT], ABC):
```
aacebo, line 19: since this is now a generic you will probably need to update the AI class and its options that contain instances of Planner

python/packages/ai/teams/ai/actions/action_types.py
```diff
@@ -17,3 +17,4 @@ class ActionTypes(str, Enum):
     PLAN_READY = "___PlanReady___"
     DO_COMMAND = "___DO___"
     SAY_COMMAND = "___SAY___"
+    TOO_MANY_STEPS = "__TooManySteps__"
```
aacebo, line 20: this new action will need to be added to the AI class with a default handler

aacebo:
you should run `poetry run fmt` to run the code formatter

python/packages/ai/teams/ai/planners/planner.py
lilyydu: good catch, thank you!

aacebo (approved):

Issue #1432, opened 2024-03-21
[Bug]: `yarn build` failed for `Teams Chef Bot` sample


MuyangAmigo:
### Language

Javascript/Typescript

### Version

latest

### Description

When executing the `yarn build` for `Teams Chef Bot` sample, it fails with unresolved dependency.
I expect I could do this from a specific sample, not from the root folder.

![image](https://github.com/microsoft/teams-ai/assets/11220663/123d23ee-a3f2-432e-bd41-29712ea462fd)

### Reproduction Steps

```bash
1. Clone the teams-ai library
2. cd to the teams chef sample
3. yarn install
4. yarn build

therealjohn:
Inside the sample directory, no `yarn build` is necessary. `yarn build` should only be run in the `js` directory when trying to create a local build of the teams-ai package to use with the sample.

dylanw-oss:
this is duplicate of #1405

dylanw-oss:
> Inside the sample directory, no `yarn build` is necessary. `yarn build` should only be run in the `js` directory when trying to create a local build of the teams-ai package to use with the sample.

@therealjohn, how about provision/deploy to Azure? I think we need to run `yarn install && yarn build` in sample folder. see #1428

corinagum:
Closing this as duplicate. @dylanw-oss I have a meeting scheduled with therealjohn to discuss the above, so in case he doesn't get to a reply I will get back to you.

Issue #1427, opened 2024-03-20
[Bug]: Can't provision sample to Azure, there is no bicep steps for openAIKey


dylanw-oss:
### Language

Javascript/Typescript

### Version

latest

### Description

This instruction does not work:
https://github.com/microsoft/teams-ai/tree/main/js/samples#deploy-the-bot-to-azure

The reason is that there is a definition of "openAIKey", but no reference in bicep file.
https://github.com/microsoft/teams-ai/blob/main/js/samples/04.ai.c.actionMapping.lightBot/infra/azure.parameters.json#L20-L22

There will be an error message when you trying to provision it:
`
(×) Error: arm/deploy failed.
    (×) Error: The ARM templates for deployment name: 'Create-resources-for-tab' could not be deployed in resource group 'bot_rg' for reason: Deployment template validation failed: 'The template parameters 'openAIKey' in the parameters file are not valid; they are not present in the original template and can therefore not be provided at deployment time. The only supported parameters for this template are 'resourceBaseName, botAadAppClientId, botAadAppClientSecret, webAppSKU, botDisplayName, serverfarmsName, webAppName, location'. Please see https://aka.ms/arm-pass-parameter-values for usage details.'.
`

### Reproduction Steps

```bash
see description
```

corinagum:
@KennethBWSong and @MuyangAmigo, could we get your assistance looking into this? Your help is appreciated!

KennethBWSong:
@corinagum In my understanding, you need to update these two parts:
1. In "infra/azure.bicep", add following line before definition of resources:
```
@secure()
param openAIKey string
```
and add following line in appSettings of resource "webApp":
![image](https://github.com/microsoft/teams-ai/assets/63089166/4dbaa8f9-fe1b-468f-942e-d1f1320ff136)
2. Create ".env.staging.user" file, and add follow line:
```
SECRET_OPENAI_KEY="your-key"
```
When running provision, ttk will read .env files and replace the placeholders in "azure.parameters.json" with values in the env variables. Then azure.bicep will consume the values in "azure.parameters.json" and provision the azure resources.

corinagum:
Lightbot and other samples have been updated to add more details about adding keys. Closing this bug as resolved.

Issue #1405, opened 2024-03-15
[Bug]: build failed due to "Messages' has no exported member 'ThreadMessage"


KennethBWSong:
### Language

Javascript/Typescript

### Version

latest

### Description

![image](https://github.com/microsoft/teams-ai/assets/63089166/d709e57c-2640-4e27-a5c5-59bff32ce8e9)

### Reproduction Steps

```bash
1.Open 06.assistants.a.mathBot template
2. npm install
3. npm run build
...
```

baishalighosh:
Running into the same issue

baishalighosh:
<img width="1046" alt="image" src="https://github.com/microsoft/teams-ai/assets/91940334/22a89a50-c759-4c49-b6b0-9426886aebab">

implodingduck:
Looks like the newer openai 4.29.1 dependency changed some exports. I had luck building by forcing openai to use version 4.28.4

implodingduck:
It was more brute forcing... 

I added the following to my "dependencies" section in package.json
```
"dependencies": {
        "@microsoft/teams-ai": "~1.1.1",
        "botbuilder": "^4.22.1",
        "botbuilder-dialogs": "^4.22.1",
        "dotenv": "^16.4.5",
        "replace": "~1.2.0",
        "restify": "~11.1.0",
        "shx": "^0.3.4",
        "openai": "4.28.4"
    }
```

then I removed my node_modules and re installed with `npm install -f`

baishalighosh:
Just tried your recommendation and this seems to work for the time being, thank you!

dylanw-oss:
run into the same issue, and I workaround it by deleting packages-lock.json and run `yarn install` and `yarn build`

KriyenKP:
Same issue, mentioned resolution does not work.

corinagum:
Hi folks, just a reminder that we use hoisting via yarn, and our directions instruct to use `yarn install && yarn build` in the root `js` folder. We can certainly work on a fix to get these bugs out of the way, but they can be avoided for now by doing the above.

baishalighosh:
[like]  Baishali Ghosh reacted to your message:
________________________________

dylanw-oss:
> It was more brute forcing...
> 
> I added the following to my "dependencies" section in package.json
> 
> ```
> "dependencies": {
>         "@microsoft/teams-ai": "~1.1.1",
>         "botbuilder": "^4.22.1",
[...]

PR #1458, opened 2024-03-27
Update samples and docs to optimize for running directly from Teams Toolkit gallery

therealjohn:
## Details

Demonstrates how a sample can be updated so it runs directly with Teams Toolkit when downloading it with the sample gallery instead of relying on guidance when contributing to the repo. 

The caveat to this approach is that it makes it slightly harder for contributors to run by requiring them to change files on first run. Since Teams Toolkit will point to samples directly in this repo, they need to run as a folder outside the repo (no Yarn steps). 

#### Change details

- Adds `npm install` step to the Teams Toolkit automation #1432 
- Adds defaults for the env files
- Updates docs / guidance
- Fixes icons to prevent validation errors
- Fixes infra files so provision and deploy to Azure works #1427 

This does work around #1405 by specifying `"openai": "4.28.4"` in the package.json.


getting-started/SAMPLES.md
```diff
@@ -11,6 +11,9 @@ _**Navigation**_
 
 ---
 
+> [!IMPORTANT]
```
corinagum, line 14: Did not know about this feature. AWESOME.

js/samples/04.ai.a.teamsChefBot/sample.env
```diff
@@ -1,5 +0,0 @@
-# This is an example file of how to set up environment variables. You can duplicate this file and add the appropriate keys.
-OPENAI_KEY=
```
corinagum, line 2: I think we still want to keep this file since dotenv is still in use. I can add a note that the env file will also need to be modified if switching to Azure OpenAI

js/samples/04.ai.a.teamsChefBot/README.md
```diff
@@ -26,31 +26,54 @@ The sample uses a local Vector Database, called [Vectra](https://github.com/Stev
...
+```json
+      "azureOpenAIKey": {
+        "value": "${{SECRET_AZURE_OPENAI_KEY}}"
+      },
+      "azureOpenAIEndpoint": {
+        "value": "${{SECRET_AZURE_OPENAI_ENDPOINT}}"
+      }
+```
+
+> Please note: If you are use Azure OpenAI, you will need both a GPT model and embedding model deployment to get this sample working. See `OpenAIModel` in `index.ts` and `OpenAIEmbeddings` in `VectraDataSource.ts`.
```
corinagum, line 74:
```suggestion
> Please note: If you are using Azure OpenAI, you will need both a GPT model and embedding model deployment to get this sample working. See `OpenAIModel` in `index.ts` and `OpenAIEmbeddings` in `VectraDataSource.ts`.
```

js/samples/04.ai.a.teamsChefBot/teamsapp.local.yml
```diff
@@ -41,10 +41,19 @@ provision:
       appPackagePath: ./build/appPackage/appPackage.${{TEAMSFX_ENV}}.zip # Relative path to this file. This is the path for built zip file.
 
 deploy:
+ # Install any dependencies and build the web app using NPM
+  - uses: cli/runNpmCommand
+    name: install dependencies
+    with:
+      args: install --no-audit
+
```
corinagum, line 49: Do we also need to have `npm run build`?

therealjohn: Not for a local run

js/samples/04.ai.a.teamsChefBot/sample.env
corinagum: the package itself would still be needed unless there's code changes. Screenshot from chefbot:
image
`config` is from `dotenv` which is what loads the `.env` contents into node's `process`:
image

singhk97: @therealjohn The issue is see with having TTK generate the .env file is that if the user wants to add a new variable to the environment, they would have to first add it to the `.env.X` file, and also add that mapping to `teamsapp.X.yml`. It makes things unnecessarily complicated in my opinion.

aacebo: yes I think we want to keep this file @therealjohn

js/samples/04.ai.a.teamsChefBot/teamsapp.local.yml
```diff
@@ -41,10 +41,19 @@ provision:
       appPackagePath: ./build/appPackage/appPackage.${{TEAMSFX_ENV}}.zip # Relative path to this file. This is the path for built zip file.
 
 deploy:
+ # Install any dependencies and build the web app using NPM
+  - uses: cli/runNpmCommand
+    name: install dependencies
+    with:
+      args: install --no-audit
```
corinagum, line 48: could you also add workspaces=false?

js/samples/04.ai.a.teamsChefBot/infra/azure.bicep
```diff
@@ -15,6 +15,15 @@ param webAppSKU string
...
 param botDisplayName string
 
+@secure()
+param openAIKey string = ''
+
+@secure()
+param azureOpenAIKey string = ''
+
+@secure()
+param auzreOpenAIEndpoint string = ''
```
corinagum, line 25:
```suggestion
param azureOpenAIEndpoint string = ''
```

js/samples/04.ai.a.teamsChefBot/sample.env
therealjohn: > @therealjohn The issue is see with having TTK generate the .env file is that if the user wants to add a new variable to the environment, they would have to first add it to the `.env.X` file, and also add that mapping to `teamsapp.X.yml`. It makes things unnecessarily complicated in my opinion.

This is necessary to handle running different environments and working well with the infra provisioning flow. 

Keeping the `.env` as it was means mixing both Teams Toolkit automation and static files, and then also mixing dotenv and env-cmd. 

I'm suggesting we pick one way - consistent with our other samples.

corinagum (approved):

singhk97 (approved):

Issue #1696, opened 2024-05-30
[Bug]: Sequence bug from citations work


corinagum:
### Language

Python

### Version

latest

### Description

error occurs in sequence apps

### Reproduction Steps

```bash
1.
2.
3.
...
```

PR #1697, opened 2024-05-30
[PY] fix: Update commands validation for sequence augmentation

corinagum:
## Linked issues

closes: #1696

### DIFF
--- a/python/packages/ai/teams/ai/augmentations/sequence_augmentation.py
+++ b/python/packages/ai/teams/ai/augmentations/sequence_augmentation.py
@@ -123,6 +123,8 @@ class SequenceAugmentation(Augmentation[Plan]):
 
             for cmd in commands:
                 if cmd["type"] == "SAY":
+                    if cmd["response"] not in cmd:
+                        cmd["response"] = None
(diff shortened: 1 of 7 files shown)


python/packages/ai/teams/ai/augmentations/sequence_augmentation.py
```diff
@@ -41,13 +41,7 @@
                     "type": {"type": "string", "enum": ["DO", "SAY"]},
                     "action": {"type": "string"},
                     "parameters": {"type": "object"},
-                    "response": {
```
lilyydu, line 44: @aacebo Corina and I were debugging and this is currently throwing an error in our test cases (its expecting type "string"- but we're still passing in the `Message` type which is an **object**, we may need to revert this code back

corinagum: image


This is the error I'm getting. 

I also tried reverting back to this schema just now and surprisingly got the same exact error message.

corinagum: To clarify: we consoled the validation response and this is that response.

lilyydu (approved):

PR #1713, opened 2024-06-05
[JS] fix: Update teamsapp.local.yml

aycabas:
#minor Unnecessary Azure AI Search key and endpoint removed from the teamsapp.local.yml that were causing error.


corinagum:
@aycabas thanks for the PR. Could you also check that the README doesn't have directions saying to update these? If they do, could you also remove those directions?

corinagum:
Hi @aycabas, the sample has undergone some changes for using managed identity, so this PR is out of date. Thank you again for the contribution.

Issue #1593, opened 2024-05-02
Rewrite OpenAI Model to use official completions client for streaming


corinagum:

Stevenic:
The `OpenAIModel` class currently uses our own custom request logic for calling OpenAI & Azure OpenAI. This is fine for completions but streaming support is a bit more complicated so I'm re-writing the class to use the official OpenAI client. This will result in a few configuration options being depreciated and replaced with new options but I'm adding logic to properly map the deprecated config options.

PR #1717, opened 2024-06-07
[JS] feat: add Streaming support 

Stevenic:
## Linked issues

closes: #1593

## Details

Updated JS `OpenAIModel` class to support streaming and updated `ActionPlanner` to support streaming chunks to the client. 

#### Change details

- Added a new `StreamingResponse` class that implements the new activity based chunking protocol.
- Added a new optional set of events that prompt completion models can implement.
- Updated the `OpenAIModel` class to implement the new prompt completion model events.
- Updated the `OpenAIModel` class to use the official node library and added streaming support.
- Added a new `StreamingLLMClient` which manages the task of streaming chunks from the model to the client using the new StreamingResponse class.
- Updated the `ActionPlanner` class to use the new `StreamingLLMClient` instead of the `LLMClient` class directly. 
- Updated the `ActionPlanner` to return an empty plan anytime streaming is used.
- Added a new `TestModel` class to simplify testing of things like streaming support.
- Added basic unit tests for the `StreamingResponse`, `TestModel`, `LLMClient`, and `StreamingLLMClient` classes. `

## Attestation Checklist

- [x ] My code follows the style guidelines of this project

  - Local testing
  - E2E testing in Teams

### Additional information

> I started basic unit tests for the `LLMClient` class. These tests should be extended to provider better coverage. We should also add unit tests for the `ActionPlanner` class now that the `TestModel` class exists.


js/packages/teams-ai/package.json
```diff
@@ -26,7 +26,7 @@
     "types": "./lib/index.d.ts",
     "peerDependencies": {
         "botbuilder": "^4.22.1",
-        "openai": "4.28.4"
+        "openai": "^4.41.0"
```
corinagum, line 29: Could you please check that all of our samples are now using v 4.41.0 as well? As a peer dependency we've added them to all of the samples.

Will there be a sample for streaming?

js/packages/teams-ai/package.json
```diff
@@ -39,6 +39,8 @@
         "gpt-tokenizer": "^2.1.2",
         "json-colorizer": "^2.2.2",
         "jsonschema": "1.4.1",
+        "openai": "^4.41.0",
+        "strict-event-emitter-types": "^2.0.0",
```
aacebo, line 43: took a look at this package, I like the idea of using event emitters but could we implement it ourselves? It would be nice to avoid adding another dependency

js/packages/teams-ai/package.json
```diff
@@ -39,6 +39,8 @@
         "gpt-tokenizer": "^2.1.2",
         "json-colorizer": "^2.2.2",
         "jsonschema": "1.4.1",
+        "openai": "^4.41.0",
```
aacebo, line 42: since this is a peer dep it doesn't need to be a regular dep

js/packages/teams-ai/src/models/index.ts
```diff
@@ -10,3 +10,4 @@ export * from './ChatCompletionAction';
 export * from './LlamaModel';
 export * from './OpenAIModel';
 export * from './PromptCompletionModel';
+export * from './TestModel';
```
aacebo, line 13: should `TestModel` be in the `internals` directory?

corinagum:
@Stevenic could you mark the class/methods in LLMClient as deprecated?

js/packages/teams-ai/src/StreamingResponse.ts
```diff
@@ -0,0 +1,129 @@
...
+            this._streamId = response?.id;
+        }
+    }   
+}
+
+/**
+ * @private
+ */
+interface StreamingChannelData {
+    streamType: 'informative' | 'streaming' | 'final';
```
lilyydu, line 126: for documentation purposes, can we note down the sequence in which these are used?

js/packages/teams-ai/package.json
Stevenic: sure and I assumed the streaming sample would happen once the Teams Client actually supports streaming.

js/packages/teams-ai/src/models/index.ts
Stevenic: I made it external so that developers could write unit tests for their bots. We could make it internal but then developers just have to re-implement the same basic logic to test their code.

js/packages/teams-ai/src/StreamingResponse.ts
Stevenic: sure...

js/packages/teams-ai/src/models/OpenAIModel.ts
```diff
@@ -39,18 +46,33 @@ export interface BaseOpenAIModelOptions {
...
 
     /**
+     * @deprecated
      * Optional. Retry policy to use when calling the OpenAI API.
      * @remarks
-     * The default retry policy is `[2000, 5000]` which means that the first retry will be after
-     * 2 seconds and the second retry will be after 5 seconds.
+     * Use `maxRetries` instead.
      */
     retryPolicy?: number[];
```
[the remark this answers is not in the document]
Stevenic: technically we can only remove things when making a major version change. This is a minor version change so I'd say no

js/packages/teams-ai/package.json
Stevenic: we could but we should probably give the original author credit somehow

Stevenic:
> @Stevenic could you mark the class/methods in LLMClient as deprecated? Notes indicating to use StreamingLLMClient instead would be great.

It's not really deprecated though. You should just generally use the derived class. It will trigger warnings if I mark it as deprecated since it's a base class

corinagum:
> It's not really deprecated though. You should just generally use the derived class. It will trigger warnings if I mark it as deprecated since it's a base class

Okay so maybe marking it as deprecated isn't the right answer, but I find it confusing that the base class is only being used by one derived class. This is specifically done for backwards compatibility, correct? In that case there should be remarks to that effect, and what scenarios you would want to use the base class. Feel free to let me know if this is just misconception from my functions oriented brain, but personally the extra comments would be beneficial.

Stevenic:
> > It's not really deprecated though. You should just generally use the derived class. It will trigger warnings if I mark it as deprecated since it's a base class
> 
> Okay so maybe marking it as deprecated isn't the right answer, but I find it confusing that the base class is only being used by one derived class. This is specifically done for backwards compatibility, correct? In that case there should be remarks to that effect, and what scenarios you would want to use the base class. Feel free to let me know if this is just misconception from my functions oriented brain, but personally the extra comments would be beneficial.

I can merge the two classes

corinagum (approved):
LGTM thanks @Stevenic!

aacebo (approved):

lilyydu (approved):

PR #1732, opened 2024-06-12
[JS] fix: Update citations sample

corinagum:
## Linked issues

#minor


js/samples/04.ai-apps/h.datasource-azureOpenAI/teamsapp.local.yml
lilyydu, line 1: should we suggest setting workspaces=true since the sample looks to be defaulted to use devtools?

corinagum: No, needing to make that change will only apply until we publish the next version. @aacebo thoughts?

js/samples/04.ai-apps/h.datasource-azureOpenAI/README.md
```diff
@@ -67,6 +67,8 @@ This sample shows how to integrate your search index as a data source into promp
 
 1. Fill the `AZURE_OPENAI_ENDPOINT`, `AZURE_SEARCH_ENDPOINT`, and `AZURE_SEARCH_INDEX` variables appropriately.
 
+1. Verify you are logged into azure cli. This is required because this sample uses managed identity. You can download and install Azure CLI from [Azure CLI docs](https://docs.microsoft.com/en-us/cli/azure/install-azure-cli). For more information on setting up environment variables, see the [Azure SDK documentation](https://github.com/Azure/azure-sdk-for-go/wiki/Set-up-Your-Environment-for-Authentication).
```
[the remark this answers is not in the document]
corinagum: mehhhh I feel like that's a given?

js/samples/04.ai-apps/h.datasource-azureOpenAI/teamsapp.local.yml
aacebo: no need to update that since by default we want the samples to use the hosted package not the local one, if we want to add it to the readme as a callout for local development instructions I could see a case for that but it would need to be added to all the sample readmes and I think that can be done in a different PR.

js/samples/04.ai-apps/h.datasource-azureOpenAI/README.md
aacebo, line 70: thanks for adding this 🙏

aacebo (approved):

PR #1832, opened 2024-07-12
[JS] Fix issues deploying to Azure with Teams Toolkit and use Managed Identity with OYD sample

therealjohn:
#minor
## Details

Removes Client ID and Client Secret from the Azure deployment and uses Managed Identity instead. 
Fixes several issues with the code running in an App Service for that deployment
Changed the default debug target to be Teams instead of Test Tool because the auth and tenant checks here won't work in the test tool. Well have to come up with a better solution for how the test tool handles this. 

```
server.post('/api/messages', async (req, res) => {
    if (!req.headers.authorization) {
        return res.send(401, 'unauthorized');
    }

    // Route received a request to adapter for processing
    await (app.adapter as TeamsAdapter).process(req, res as any, async (context: any) => {
        if (context.activity.conversation.tenantId !== process.env.BOT_TENANT_ID) {
            return res.send(401, 'invalid tenant');
        }

        // Dispatch to application for routing
        await app.run(context);
    });
});
```


js/samples/04.ai-apps/h.datasource-azureOpenAI/src/app.ts
```diff
@@ -48,20 +48,21 @@ export const app = new Application<ApplicationTurnState>({
...
 
-app.conversationUpdate('membersAdded', async (context) => {
+app.conversationUpdate('membersAdded', async (context: TurnContext) => {
     await context.sendActivity(
         "Welcome! I'm a conversational bot that can tell you about your data. You can also type `/clear` to clear the conversation history."
     );
 });
 
-app.message('/clear', async (context, state) => {
+app.message('/clear', async (context: TurnContext, state: any) => {
```
aacebo, line 65: state type should be some variant of `TurnState`

js/samples/04.ai-apps/h.datasource-azureOpenAI/src/app.ts
```diff
@@ -92,7 +93,7 @@ app.error(async (context: TurnContext, err: any) => {
     await context.sendActivity('To continue to run this bot, please fix the bot source code.');
 });
 
-app.feedbackLoop(async (_context, _state, feedbackLoopData) => {
+app.feedbackLoop(async (_context: TurnContext, _state: any, feedbackLoopData: any) => {
```
aacebo, line 96: same, avoid using `any`

js/samples/04.ai-apps/h.datasource-azureOpenAI/src/index.ts
```diff
@@ -28,13 +28,13 @@ server.listen(port, () => {
...
 server.post('/api/messages', async (req, res) => {
-    if (!req.headers.authorization) {
+    if (!req.headers.authorization && process.env.RUNNING_ON_AZURE === "1") {
         return res.send(401, 'unauthorized');
     }
 
     // Route received a request to adapter for processing
-    await (app.adapter as TeamsAdapter).process(req, res as any, async (context) => {
-        if (context.activity.conversation.tenantId !== process.env.BOT_TENANT_ID) {
+    await (app.adapter as TeamsAdapter).process(req, res as any, async (context: any) => {
```
aacebo, line 36: same, avoid using `any`

js/samples/04.ai-apps/h.datasource-azureOpenAI/teamsapp.local.yml
```diff
@@ -54,5 +65,5 @@ deploy:
               BOT_ID: ${{BOT_ID}}
               BOT_PASSWORD: ${{SECRET_BOT_PASSWORD}}
               BOT_TENANT_ID: ${{TEAMS_APP_TENANT_ID}}
-              AZURE_OPENAI_ENDPOINT: ${{AZURE_OPENAI_ENDPOINT}}
```
aacebo, line 57: should this be passing in `AZURE_OPENAI_DEPLOYMENT` now?

js/samples/04.ai-apps/h.datasource-azureOpenAI/env/.env.local
```diff
@@ -4,5 +4,4 @@ TEAMSFX_ENV=local
 APP_NAME_SUFFIX=local
 TEAMS_APP_ID=
 TEAMS_APP_TENANT_ID=
-BOT_ID=
-AZURE_OPENAI_ENDPOINT=
```
aacebo, line 8: is `AZURE_OPENAI_ENDPOINT` being removed? I still see references to it in the code

aacebo (changes requested):

js/samples/04.ai-apps/h.datasource-azureOpenAI/env/.env.local.user
```diff
@@ -1,2 +1 @@
 SECRET_BOT_PASSWORD=
```
corinagum, line 1: (applies to file below) -- does .`env.staging` only need to be removed for this sample because of managed identity? Or should we be updating the other samples as well?

Carter425 (commented):
https://github.com/microsoft/teams-ai/pull/1832/commits/fe7360cbe209e74b0ce253c97351c923199c28fa

Given John's changes on 7-12 that enables the Test Tool, the default testing method should remain Debug in Test Tool. The specific changes to default to Debug in Teams (Edge) are no longer needed nor should be implemented as the Test Tool is the preferred testing method.

js/samples/04.ai-apps/h.datasource-azureOpenAI/teamsapp.local.yml
therealjohn: @Carter425 I added this - we should put this in the instructions. The sample had it hard-coded, but its a reasonable thing that you would paste along with your endpoint.

js/samples/04.ai-apps/h.datasource-azureOpenAI/env/.env.local
therealjohn: Its not removed, but it already exists in the sample.env which the instructions point you to so its not needed here.

js/samples/04.ai-apps/h.datasource-azureOpenAI/env/.env.local.user
therealjohn: I removed it because I didn't recognize it - I thought it was a mistake. If its needed let me know. 

`dev` serves the same purpose as a default remote environment.

corinagum: Oh interesting, we've been adding it to all our samples and this whole time I didn't know the purpose :D

js/samples/04.ai-apps/h.datasource-azureOpenAI/teamsapp.local.yml
Carter425: @therealjohn what is the change that needs to be made to the instructions?

aacebo (approved):

aacebo (approved):

PR #1839, opened 2024-07-15
[PY] Copy activity to new activity when continuing conversation

skoryky:
When `Application` is instantiated with `long_running_messages=True`, the incoming request is converted to a proactive conversation, which creates a new activity. The original code has a bug where the original activity's attributes aren't copied to the new activity, which results in the original activity intent not being handled. Fix this by copying the activity in a new callback function which wraps the original callback function.

### DIFF
--- a/python/packages/ai/teams/app.py
+++ b/python/packages/ai/teams/app.py
@@ -806,9 +806,14 @@ class Application(Bot, Generic[StateT]):
             and ActivityTypes.message == context.activity.type
             and self._options.long_running_messages
         ):
+            def wrapped_callback(ctx: TurnContext) -> Awaitable:
+                for key in context.activity._attribute_map.keys():
+                    setattr(ctx.activity, key, getattr(context.activity, key))
+                return func(ctx)
+
             return await self._adapter.continue_conversation(
                 reference=context.get_conversation_reference(context.activity),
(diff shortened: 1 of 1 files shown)


python/packages/ai/teams/app.py
```diff
@@ -806,9 +806,14 @@ async def _start_long_running_call(
             and ActivityTypes.message == context.activity.type
             and self._options.long_running_messages
         ):
+            def wrapped_callback(ctx: TurnContext) -> Awaitable:
```
aacebo, line 809: if the new activity is not being created using the old activity that would be a bug in the botframework

skoryky: I copied this logic from the [JS library code](https://github.com/microsoft/teams-ai/blob/main/js/packages/teams-ai/src/Application.ts#L1060-L1063), which made me think it's a bug in this library.

aacebo: im following up with botframework on this, will get back to you with an answer.

aacebo: I followed up with BF and they said there is no need to copy the activity over

skoryky: Do you mean they're fixing something on their end?

PR #1842, opened 2024-07-15
[PY] fix: Fix Application's activity selector to not cast ActivityTypes to str

skoryky:
`ActivityTypes` inherits from `str` (in addition to `Enum`), so does not need to be cast to `str` before being compared to `context.activity.type`. Casting to `str` will convert a value to something like `ActivityTypes.message`, which will fail the selector comparison.

#minor

### DIFF
--- a/python/packages/ai/teams/app.py
+++ b/python/packages/ai/teams/app.py
@@ -211,7 +211,7 @@ class Application(Bot, Generic[StateT]):
         """
 
         def __selector__(context: TurnContext):
-            return type == str(context.activity.type)
+            return type == context.activity.type
 
         def __call__(func: RouteHandler[StateT]) -> RouteHandler[StateT]:
             self._routes.append(Route[StateT](__selector__, func))


python/packages/ai/teams/app.py
```diff
@@ -211,7 +211,7 @@ async def on_event(context: TurnContext, state: TurnState):
         """
 
         def __selector__(context: TurnContext):
-            return type == str(context.activity.type)
```
aacebo, line 214: we need this cast because the bot framework type for `context.activity.type` is `Unknown | None`, because they did not add type annotations.

skoryky: Ah, I see! Looks like none of the other selectors in this file cast to `str`. What's the reason for that?

Would it be better to do:

```
if not context.activity.type:
    return False

if context.activity.type and isinstance(context.activity.type, ActivityTypes):
    return type == context.activity.type
else:
    return type == str(context.activity.type)
```

aacebo: just checked locally and it looks like `mypy` isn't throwing any type errors when the cast is removed, so we can update this to match the other selectors. Sorry for the confusion.

aacebo (approved):

PR #1830, opened 2024-07-10
[PY] fix: Use security code stored in state param to get user token

skoryky:
## Linked issues

closes: # (issue number)

## Details

The OAuth get_token() function does not use the security code stored in the state param to look up the user token stored when the user signed in earlier. This results in an infinite loop of sign-in prompts when calling Auth's sign_in().

#### Change details

> Describe your changes, with screenshots and code snippets as appropriate

**code snippets**:

**screenshots**:

## Attestation Checklist

- [ ] My code follows the style guidelines of this project

- I have checked for/fixed spelling, linting, and other errors
- I have commented my code for clarity
- I have made corresponding changes to the documentation (updating the doc strings in the code is sufficient)
- My changes generate no new warnings
- I have added tests that validates my changes, and provides sufficient test coverage. I have tested with:
  - Local testing
  - E2E testing in Teams
- New and existing unit tests pass locally with my changes

### Additional information

> Feel free to add other relevant information below

PR #1843, opened 2024-07-15
[PY] Avoid returning token prematurely on signin/verifyState invoke

skoryky:
The original code skips calling `auth.verify_state` on sign-in. This fix is only a band-aid, as there is no handler registered to process `signin/verifyState` invokes, so these requests ultimately end up receiving 501 Not Implemented responses.

### DIFF
--- a/python/packages/ai/teams/auth/auth_manager.py
+++ b/python/packages/ai/teams/auth/auth_manager.py
@@ -65,7 +65,7 @@ class AuthManager(Generic[StateT]):
         token: Optional[str] = await auth.get_token(context)
         res = SignInResponse("pending")
 
-        if token:
+        if token and not self._is_verify_state_activity(context):
             cast(TempState, state.temp).auth_tokens[key] = token
             return SignInResponse("complete")
 


python/packages/ai/teams/auth/auth_manager.py
```diff
@@ -65,7 +65,7 @@ async def sign_in(
         token: Optional[str] = await auth.get_token(context)
         res = SignInResponse("pending")
 
-        if token:
+        if token and not self._is_verify_state_activity(context):
```
aacebo, line 68: if its a verify state activity there should not be a token present correct? I'm confused why this check is needed.

skoryky: `_is_verify_state_activity` returns True when the activity name is `signin/verifyState`. This activity happens after sign in has succeed, and we now need to call `verify_state` [here](https://github.com/microsoft/teams-ai/blob/main/python/packages/ai/teams/auth/auth_manager.py#L95-L96), according to [this comment](https://github.com/microsoft/teams-ai/blob/main/python/packages/ai/teams/auth/auth.py#L81-L83) in the code that indicates `verify_state` needs to be called on `signin/verifyState`.

aacebo: yes but if the activity is `signin/verifyState` then there should not be a token present, because the point of the `verify_state` activity/method is to exchange a state variable for a token response.

skoryky: Ohh, I see...

When going through the OAuth2 sign-in flow through a bot, a `signin/verifyState` message is sent. However, when going through the sign-in flow when composing a message extension, such a message is not sent. This previously resulted in an infinite loop of sign-in prompts, which I tried to fix in [this PR](https://github.com/microsoft/teams-ai/pull/1830) [1].

However, that change resulted in the token being available [here](https://github.com/microsoft/teams-ai/blob/1282a0757d2217905a3ed6a0ec359e74a5d75c0b/python/packages/ai/teams/auth/auth_manager.py#L65-L68) when processing a `signin/verifyState` message, such that the function returns early and does not call the `_on_sign_in_success` callback [farther down in the function](https://github.com/microsoft/teams-ai/blob/1282a0757d2217905a3ed6a0ec359e74a5d75c0b/python/packages/ai/teams/auth/auth_manager.py#L114-L115). Because of this, the `signin/verifyState` message shows as failed on the client. That's what I was trying to fix in the current PR.

Why isn't a `signin/verifyState` sent for message extensions? I think when using the Bot Framework SDK, you'd be expected to complete the authentication and retrieve the token in the `composeExtension/query` handler. But how should this be done with Teams AI library when using `auto=True` on `AuthOptions`? There's no opportunity to get the token using state before the sign-in flow is initiated again, and there's no `signin/verifyState` message to do it for you. Should it be done somewhere in `Application._authenticate_user` ([link](https://github.com/microsoft/teams-ai/blob/1282a0757d2217905a3ed6a0ec359e74a5d75c0b/python/packages/ai/teams/app.py#L716))?

[1] Should we roll this change back?

skoryky: Bumping this thread. I'm now needing to set `auto=False` on `AuthOptions` and doing auth manually. However, the `signin/verifyState` message isn't handled anywhere by this library. How is this supposed to work? Should there be a handler for that message in `app.py`?

Issue #1777, opened 2024-06-26
[Bug]: If you send a message without any text and only an image, there is an exception


heyitsaamir:
### Language

Python

### Version

latest

### Description

The bug is here - https://github.com/microsoft/teams-ai/blob/2b787dbd5d923df0a6b0ea23b236686afdd83220/python/packages/ai/teams/app.py#L253

If you paste an image and send it, you get an exception that looks like this:
```terminal
 [on_turn_error] unhandled error: 'NoneType' object has no attribute 'find'
Traceback (most recent call last):
  File "workspace/.venv/lib/python3.11/site-packages/botbuilder/core/bot_adapter.py", line 174, in run_pipeline
    return await self._middleware.receive_activity_with_status(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "workspace/.venv/lib/python3.11/site-packages/botbuilder/core/middleware_set.py", line 69, in receive_activity_with_status
    return await self.receive_activity_internal(context, callback)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "workspace/.venv/lib/python3.11/site-packages/botbuilder/core/middleware_set.py", line 79, in receive_activity_internal
    return await callback(context)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "workspace/.venv/lib/python3.11/site-packages/teams/app.py", line 663, in on_turn
    await self._start_long_running_call(context, self._on_turn)
  File "workspace/.venv/lib/python3.11/site-packages/teams/app.py", line 813, in _start_long_running_call
    return await func(context)
           ^^^^^^^^^^^^^^^^^^^
  File "workspace/.venv/lib/python3.11/site-packages/teams/app.py", line 742, in _on_turn
    is_ok, matches = await self._on_activity(context, state)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "workspace/.venv/lib/python3.11/site-packages/teams/app.py", line 790, in _on_activity
    for route in routes:
  File "workspace/.venv/lib/python3.11/site-packages/teams/app.py", line 780, in <lambda>
    routes = filter(lambda r: not r.is_invoke and r.selector(context), self._routes)
                                                  ^^^^^^^^^^^^^^^^^^^
[...]

PR #1850, opened 2024-07-17
[PY] fix: Do not throw error if there is no text content (but attachments)

heyitsaamir:
## Linked issues

closes: #1777

## Details

If there is a message that has no text content, but it does have some non-textual attachments, we currently throw. However, the message could have attachments that could be used for further processing by the pipeline. As such, it doesn't make sense to throw an error in this case.

#### Change details

Before the error:
image

After the error:

https://github.com/user-attachments/assets/01a1f260-8bde-4d01-bd55-a660c1a905fa

Note: 
I didn't see any existing tests that tested this particular path (esp using `options.ai`). Any pointers on building that out? I started adding to `test_app.py` but wanted to know if there were easy ways of mocking out the AI layer out. Hence no tests yet, but happy to add it if it's not that hard.

## Attestation Checklist

  - Local testing
  - E2E testing in Teams

### Additional information


lilyydu (approved):
Appreciate the find and fix! Mocking out the AI portion is a bit more complex (we haven't gotten to it yet) and the check is also private so good on tests for now

corinagum:
This would need to be added to JS and C# as well

Issue #1699, opened 2024-05-30
[PY]: Add tools support to OAI models


lilyydu:
- primarily for Chat Completions

PR #1853, opened 2024-07-17
[PY] feat: Chat Completion - Tools

lilyydu:
## Linked issues

closes: #1699

## Details

- providing feature support w/ function calling in [chat completions](https://platform.openai.com/docs/guides/function-calling)

## Attestation Checklist

  - Local testing
  - E2E testing in Teams


python/packages/ai/teams/ai/augmentations/tools_augmentation.py
```diff
@@ -0,0 +1,198 @@
...
+            memory (MemoryBase): Interface for accessing state variables.
+            response (PromptResponse[Union[str, List[chat.ChatCompletionMessageToolCall]]]):
+                The validated and transformed response for the prompt.
+
+        Returns:
+            Plan: The created plan.
+        """
+
+        tool_map: Dict = {}
+        commands: List[Union[PredictedDoCommand, PredictedSayCommand]] = []
```
corinagum, line 178: Typescript has a type called `PredictedCommand` which is already a Union type -- you should be able to use that instead of defining a new Union type here.

python/packages/ai/teams/ai/planners/action_planner.py
```diff
@@ -149,6 +158,43 @@ async def complete_prompt(
             template=template,
         )
 
+    def _handle_action_tools(self, memory: MemoryBase, template: PromptTemplate) -> MemoryBase:
+        if not memory.get(SUBMIT_TOOL_OUTPUTS_VARIABLE):
+            memory.set("temp.tool_choice", template.config.completion.tool_choice)
+            memory.set("temp.parallel_tool_calls", template.config.completion.parallel_tool_calls)
```
corinagum, line 164: Do these need to be set here if we're resetting the state immediately below?

corinagum: try to call tools - couldn't validate
first time calling tools 

first time setting them to memory (technically could be repeated, see above)

python/packages/ai/teams/ai/augmentations/tools_augmentation.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+class ToolsAugmentation(Augmentation[Union[str, List[chat.ChatCompletionMessageToolCall]]]):
+    """
+    A server-side 'tools' augmentation.
+    """
+
+    _actions: List[ChatCompletionAction] = []
+
+    def __init__(self, actions: Optional[List[ChatCompletionAction]] = None) -> None:
```
aacebo, line 38: NIT: instead of using an optional and an if statement here, it would be cleaner for the argument to default to an empty list so you can just assign the variable without an if statement.

python/packages/ai/teams/ai/augmentations/tools_augmentation.py
```diff
@@ -0,0 +1,198 @@
...
+        if actions:
+            self._actions = actions
+
+    def create_prompt_section(self) -> Union[PromptSection, None]:
```
aacebo, line 42: this turn type should be `Optional[PromptSection]`

python/packages/ai/teams/ai/augmentations/tools_augmentation.py
```diff
@@ -0,0 +1,198 @@
...
+                                valid_action_tool_calls.append(curr_tool_call)
+                        else:
+                            valid_action_tool_calls.append(curr_tool_call)
+                    else:
+                        return Validation(valid=False, feedback="The invoked tool does not exist.")
+
+                # Calling multiple tools
+                else:
+                    for tool_call in tool_calls:
+                        function_name = tool_call.function.name
```
aacebo, line 117: instead of duplicating all this logic between the if and the else, you can use the if and the else to initialize an array of tool calls and put this logic into a single look which would result in probably half the lines of code.

python/packages/ai/teams/ai/augmentations/tools_constants.py
```diff
@@ -0,0 +1,20 @@
+"""
+Copyright (c) Microsoft Corporation. All rights reserved.
+Licensed under the MIT License.
+"""
+
+from __future__ import annotations
+
+# Set of constants used for action tool handling
+
+SUBMIT_TOOL_OUTPUTS_VARIABLE = "temp.submit_tool_outputs"
```
aacebo, line 10: I don't think this is needed, you should be able to just check your `SUBMIT_TOOL_OUTPUTS_MAP` and if it has a length > 0 then you have outputs

python/packages/ai/teams/ai/augmentations/tools_constants.py
```diff
@@ -0,0 +1,20 @@
...
+# Boolean flag to denote roundtrip tool calls
+
+SUBMIT_TOOL_OUTPUTS_MAP = "temp.submit_tool_map"
+# Map of tool_name to tool_id
+
+SUBMIT_TOOL_OUTPUTS_MESSAGES = "temp.submit_tool_messages"
+# List of tool_output messages
+
+SUBMIT_TOOL_HISTORY = "temp.tool_history"
```
aacebo, line 19: are we now storing conversation history in the `conversation` state and in the `temp` state? I think we should find a way to leverage the existing history state instead.

python/packages/ai/teams/ai/augmentations/tools_constants.py
```diff
@@ -0,0 +1,20 @@
...
+
+# Set of constants used for action tool handling
+
+SUBMIT_TOOL_OUTPUTS_VARIABLE = "temp.submit_tool_outputs"
+# Boolean flag to denote roundtrip tool calls
+
+SUBMIT_TOOL_OUTPUTS_MAP = "temp.submit_tool_map"
+# Map of tool_name to tool_id
+
+SUBMIT_TOOL_OUTPUTS_MESSAGES = "temp.submit_tool_messages"
```
aacebo, line 16: should should be able to get this from the existing conversation history state

python/packages/ai/teams/ai/models/prompt_response.py
```diff
@@ -11,7 +11,9 @@
 from ..prompts.message import Message
 
 ContentT = TypeVar("ContentT")
-PromptResponseStatus = Literal["success", "error", "rate_limited", "invalid_response", "too_long"]
+PromptResponseStatus = Literal[
+    "success", "error", "rate_limited", "invalid_response", "too_long", "tools_error"
```
aacebo, line 15: we should not use `tools` terminology` when possible since its a concept of openai and not necessarily all language models. Is there a reason we can't reuse `error` ?

python/packages/ai/teams/ai/planners/action_planner.py
```diff
@@ -149,6 +158,43 @@ async def complete_prompt(
             template=template,
         )
 
+    def _handle_action_tools(self, memory: MemoryBase, template: PromptTemplate) -> MemoryBase:
+        if not memory.get(SUBMIT_TOOL_OUTPUTS_VARIABLE):
+            memory.set("temp.tool_choice", template.config.completion.tool_choice)
```
aacebo, line 163: why do these state variables not have constants like the others?

python/packages/ai/teams/ai/prompts/message.py
```diff
@@ -26,13 +27,16 @@ class Message(Generic[T], DataClassJsonMixin):
...
+        action_tool_calls (Optional[List[chat.ChatCompletionMessageToolCall]]):
+            List of tool calls from the model.
     """
 
     role: str
     content: Optional[T] = None
     context: Optional[MessageContext] = None
     function_call: Optional[FunctionCall] = None
     name: Optional[str] = None
+    action_tool_calls: Optional[List[chat.ChatCompletionMessageToolCall]] = None
```
aacebo, line 39: this is our message type so to keep things generic in types like this the name `action_calls` and mapping to our own internal type, not an openai type, helps us keep this code reusable with different clients/types/models.

python/packages/ai/teams/ai/prompts/completion_config.py
```diff
@@ -58,6 +60,15 @@ class CompletionConfig:
         top_p (float): The model's top_p as a number between 0 and 2.
           Defaults to 0.
 
+        include_tools (Optional[bool]): If True, function calling will be enabled with the LLM.
+          Defaults to False.
+
+        tool_choice (Optional[chat.ChatCompletionToolChoiceOptionParam]):
```
aacebo, line 66: I don't think we need to include these, since we are using `actions.json` as a way to map our actions to send to the model.

python/packages/ai/teams/ai/augmentations/tools_augmentation.py
corinagum: `createPromptSection`/ `create_prompt_section` in the other augmentations matches the pattern of returning `PromptSection` or `undefined`, so I think it's okay as-is. Also this method isn't performing/returning anything, like in `DefaultAugmentation`/`default_augmentation`

python/packages/ai/teams/ai/prompts/completion_config.py
corinagum: `tool_choice` is one of the [params](https://platform.openai.com/docs/api-reference/chat/create#chat-create-tool_choice) sent to the API call, so we want the customer to be able to define this. Did you by chance mean `include_tools`? That we are using to allow enabling/disabling tools, but with the logic being put into `ToolsAugmentation` instead of the model, I can see why that might be unnecessary - if they have actions and it's a tools augmentation, we can assume they want to use server-side tools. Thoughts?

python/packages/ai/teams/ai/augmentations/tools_augmentation.py
aacebo: `Optional[T]` is the same as `Union[T, None]` https://docs.python.org/3/library/typing.html#typing.Optional

we should also update the augmentation abstract class to reflect that https://github.com/microsoft/teams-ai/blob/main/python/packages/ai/teams/ai/augmentations/augmentation.py#L30

corinagum: My bad - wasn't think of this as a python-specific nit. Thanks for clarifying.

python/packages/ai/teams/ai/prompts/completion_config.py
aacebo: yes I meant for `include_tools` which is not part of the api. enabling/disabling of tools functionality whould be configured by using `augmentation_type: "none"` instead of tools augmentation.

python/packages/ai/teams/ai/prompts/message.py
corinagum: here's the interface for `ChatCompletionMessageToolCall`:
```ts
export interface ChatCompletionMessageToolCall {
  /**
   * The ID of the tool call.
   */
  id: string;

  /**
   * The function that the model called.
   */
  function: ChatCompletionMessageToolCall.Function;

  /**
   * The type of the tool. Currently, only `function` is supported.
   */
  type: 'function';
```
which imo doesn't map well enough with `FunctionCall`:
```ts
export interface FunctionCall {
    /**
     * Name of the function to call.
     */
    name?: string;

    /**
     * Optional. Arguments to pass to the function. Must be deserialized.
     */
    arguments?: string;
}
```
fThere's also `ChatCompletionAction` defined in our library, but that doesn't map well either. 
So I was thinking of making a new interface. Thoughts?
@aacebo @Stevenic

python/packages/ai/teams/ai/augmentations/tools_constants.py
lilyydu: the map doesn't get populated until we hit `tools_augmentation.create_plan_from_response` - but before it reaches this code, we use this var to check if the model requested to call tools

python/packages/ai/teams/ai/augmentations/tools_constants.py
lilyydu: we're storing the messages we're sending to/from the model here, but open to leveraging history state instead.

how are we leveraging `state.conversation` right now? I only see it in the `dialogs` logic (`state.conversation[self.id]`) - is this used with the `ConversationHistorySection `class?

python/packages/ai/teams/ai/prompts/message.py
lilyydu: I think it's best to create our own type, potentially called `ActionCall` that looks somewhat similiar to `ChatCompletionMessageToolCall`.

python/packages/ai/teams/ai/augmentations/tools_constants.py
aacebo: right now history is persisted using this variable, which your code should use to get all the data it needs. 

https://github.com/microsoft/teams-ai/blob/main/python/packages/ai/teams/ai/planners/action_planner.py#L135
https://github.com/microsoft/teams-ai/blob/main/python/packages/ai/teams/ai/clients/llm_client.py#L37

python/packages/ai/teams/ai/prompts/message.py
aacebo: yes, I agree

python/packages/ai/teams/ai/augmentations/tools_constants.py
aacebo: you should be able to use the conversation history to determine this, what you are trying to determine is if there are any tool calls that have not yet been executed in the conversation history

python/packages/ai/teams/ai/planners/action_planner.py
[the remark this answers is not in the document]
lilyydu: TMK we don't have access to the `template` inside tools augmentation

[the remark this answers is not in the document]
corinagum: wow you're totally right, sorry about that. I even checked and apparently looked at the wrong file. thanks Lily~

python/packages/ai/teams/ai/augmentations/__init__.py
```diff
@@ -7,10 +7,16 @@
...
+from .tools_augmentation import ToolsAugmentation
+from .tools_constants import ACTIONS_HISTORY, PARALLEL_TOOL_CALLS, TOOL_CHOICE
 
 __all__ = [
     "Augmentation",
     "DefaultAugmentation",
     "MonologueAugmentation",
     "SequenceAugmentation",
+    "ToolsAugmentation",
+    "ACTIONS_HISTORY",
```
aacebo, line 19: are these for internal use only or are we exposing them for devs? If they are internal only they should not be exported from the `__init__.py`

python/packages/ai/teams/ai/augmentations/tools_augmentation.py
```diff
@@ -0,0 +1,189 @@
...
+
+class ToolsAugmentation(Augmentation[Union[str, List[ActionCall]]]):
+    """
+    A server-side 'tools' augmentation.
+    """
+
+    _actions: List[ChatCompletionAction] = []
+
+    def __init__(self, actions: List[ChatCompletionAction] = []) -> None:
+        # pylint: disable=dangerous-default-value
```
aacebo, line 38: its best to either disable the pylint rule if it is not helpful, or find a way to satisfy it with your code. Inline disables are almost never a good idea

python/packages/ai/teams/ai/augmentations/tools_augmentation.py
```diff
@@ -0,0 +1,189 @@
...
+            response.message
+            and response.message.action_calls
+            and len(response.message.action_calls) > 0
+            and memory.has(ACTIONS_HISTORY)
+        ):
+            tool_calls = response.message.action_calls
+            tools = self._actions
+
+            if tools and len(tool_calls) > 0:
+                tool_choice = memory.get(TOOL_CHOICE) or "auto"
```
aacebo, line 81: shouldn't `tool_choice` be set in the `config.json` by the developer?

python/packages/ai/teams/ai/models/openai_model.py
```diff
@@ -124,11 +125,47 @@ async def complete_prompt(
         template: PromptTemplate,
     ) -> PromptResponse[str]:
         max_tokens = template.config.completion.max_input_tokens
+
+        # Setup actions if enabled
+        is_tools_aug = (
+            template.config.augmentation
+            and template.config.augmentation.augmentation_type == "tools"
+        )
+        tool_choice = (
```
aacebo, line 134: im confused by this being configurable in the `config.json` but also a temp state variable.

python/packages/ai/teams/ai/models/openai_model.py
```diff
@@ -151,6 +188,14 @@ async def complete_prompt(
...
 
+        # Submit tool outputs, if previously invoked
+        if memory.has(ACTIONS_HISTORY):
+            tools_messages = cast(
+                List[chat.ChatCompletionMessageParam], memory.get(ACTIONS_HISTORY)
+            )
+            messages.extend(tools_messages)
+            memory.set(ACTIONS_HISTORY, [])
+
         for msg in res.output:
```
aacebo, line 199: doesn't this for loop need to be updated to support tool messages?

python/packages/ai/teams/ai/models/openai_model.py
```diff
@@ -195,12 +241,60 @@ async def complete_prompt(
...
                 self._options.logger.debug("COMPLETION:\n%s", completion.model_dump_json())
 
+            # Handle tools flow
+            action_calls: List[ActionCall] = []
+            response_message = completion.choices[0].message
+
+            if is_tools_aug:
+                tool_calls = response_message.tool_calls
+                if tool_calls and len(tool_calls) > 0:
+                    if not parallel_tool_calls and len(tool_calls) > 1:
```
aacebo, line 260: you don't need to validate the responses from the openai service, you can assume the service will return valid responses.

python/packages/ai/teams/ai/planners/action_planner.py
```diff
@@ -149,6 +157,45 @@ async def complete_prompt(
...
+                    curr_message.tool_calls or [] # type: ignore[attr-defined]
+                )
+                tool_outputs: List[chat.ChatCompletionToolMessageParam] = []
+
+                for action in action_outputs:
+                    output = action_outputs[action]
+                    tool_call_id = None
+
+                    for tool in tools:
+                        if tool.function.name == action: # type: ignore[attr-defined]
```
aacebo, line 180: better to use `getattr` instead, never use type ignores or lint ignores inline

python/packages/ai/teams/ai/models/openai_model.py
lilyydu: yup it's defined by the developer in the config.json but we have it saved in temp state bc we don't have access to the template in tools_augmentation

python/packages/ai/teams/ai/models/openai_model.py
lilyydu: nope, we're propagating it over in `L192-197` via the changes in action_planner

python/packages/ai/teams/ai/augmentations/tools_augmentation.py
```diff
@@ -0,0 +1,115 @@
...
+            tool_calls = response.message.action_calls
+            tools = self._actions
+
+            if tools and len(tool_calls) > 0:
+                return Validation(
+                    valid=True,
+                    value=tool_calls,
+                )
+
+        memory.set(ACTIONS_HISTORY, [])
```
aacebo, line 78: whenever the LLM responds you clear the conversation history?

python/packages/ai/teams/ai/models/openai_model.py
```diff
@@ -124,11 +125,47 @@ async def complete_prompt(
...
+        parallel_tool_calls = (
+            template.config.completion.parallel_tool_calls
+            if template.config.completion.parallel_tool_calls is not None
+            else True
+        )
+        tools: List[chat.ChatCompletionToolParam] = []
+
+        # If tools is enabled, reformat actions to schema
+        if is_tools_aug:
+            if not template.actions or len(template.actions) == 0:
```
aacebo, line 148: this should not throw an exception, you can use tools while not providing any tools without the api throwing an error

python/packages/ai/teams/ai/models/openai_model.py
```diff
@@ -195,12 +241,43 @@ async def complete_prompt(
...
+                tool_calls = response_message.tool_calls
+                if tool_calls and len(tool_calls) > 0:
+                    messages.append(
+                        cast(chat.ChatCompletionAssistantMessageParam, response_message)
+                    )
+
+                    if not memory.has("conversation"):
+                        memory.set("conversation", {})
+
+                    memory.set(ACTIONS_HISTORY, messages)
```
aacebo, line 267: you should not ever need to set the history yourself, since that logic is being handled by the `LLMClient` https://github.com/microsoft/teams-ai/blob/lilyydu/tools/python/packages/ai/teams/ai/planners/action_planner.py#L140 https://github.com/microsoft/teams-ai/blob/lilyydu/tools/python/packages/ai/teams/ai/clients/llm_client.py#L218

python/packages/ai/teams/ai/augmentations/tools_augmentation.py
lilyydu: cleared if the LLM responds with no tools (for sanity purposes to ensure a clean slate for subsequent calls)

python/packages/ai/teams/ai/models/openai_model.py
lilyydu: what would be the purpose of using tools aug without specifying any tools in `actions.json`? or do you mean the onus is on the developer to configure correctly

python/packages/ai/teams/ai/augmentations/tools_augmentation.py
```diff
@@ -0,0 +1,100 @@
...
+
+        Returns:
+            Validation: A 'Validation' object.
+        """
+        return Validation(valid=True)
+
+    async def create_plan_from_response(
+        self,
+        turn_context: TurnContext,
+        memory: MemoryBase,
```
corinagum, line 70: Is there a reason for using `MemoryBase` instead of `Memory`? In .ts, I see the augmentation files using `Memory`.

python/packages/ai/teams/ai/augmentations/tools_augmentation.py
```diff
@@ -0,0 +1,100 @@
...
+from ..validators.validation import Validation
+from .augmentation import Augmentation
+
+
+class ToolsAugmentation(Augmentation[str]):
+    """
+    A server-side 'tools' augmentation.
+    """
+
+    _actions: Optional[List[ChatCompletionAction]]
```
corinagum, line 33: Do we still need this?

python/packages/ai/teams/ai/clients/llm_client.py
```diff
@@ -91,27 +90,19 @@ def __init__(self, options: LLMClientOptions) -> None:
 
         self._options = options
 
-    def add_function_result_to_history(self, memory: MemoryBase, name: str, results: Any) -> None:
+    def add_action_output_to_history(self, memory: MemoryBase, id: str, results: str) -> None:
```
corinagum, line 93: Not sure if we can delete `add_function_result_to_history` since it's public. @aacebo thoughts? I suspect it was meant as a helper function for customers and since it's public we can't delete it ad-hoc.

python/packages/ai/teams/ai/augmentations/tools_augmentation.py
lilyydu: good catch, will remove

python/packages/ai/teams/ai/clients/llm_client.py
lilyydu: I did a reference check and we're not using it anywhere internally but it is a fair concern. Happy to just add it as another function if thats the case

aacebo: it is a public function but the llmclient isn't a public class (we don't even reference it in our docs) so its fine to remove.

aacebo (approved):
LGTM!!!

PR #1860, opened 2024-07-19
[PY] fix: Match against text only if it exists in message selector

skoryky:
## Linked issues

#minor

## Details

When a message extension is inserted into a chat with a bot, a request with type `message` is sent with no `text` field, resulting in the following error:

`AttributeError: 'NoneType' object has no attribute 'find'`

Check for the existence of `text` before calling `find` on it.

### DIFF
--- a/python/packages/ai/teams/app.py
+++ b/python/packages/ai/teams/app.py
@@ -251,8 +251,11 @@ class Application(Bot, Generic[StateT]):
                 hits = re.match(select, text)
                 return hits is not None
 
-            i = context.activity.text.find(select)
-            return i > -1
+            if context.activity.text:
+                i = context.activity.text.find(select)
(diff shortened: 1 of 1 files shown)


python/packages/ai/teams/app.py
```diff
@@ -251,8 +251,11 @@ def __selector__(context: TurnContext):
                 hits = re.match(select, text)
                 return hits is not None
 
-            i = context.activity.text.find(select)
```
lilyydu, line 254: Hi @skoryky, we just had another fix for this via #1850- can you check if you are still getting this error with ME's?

skoryky: Awesome, that other PR takes care of it. Closing!

Issue #1920, opened 2024-08-08
[JS] Streaming - Test


lilyydu:
[Pending updates in JS]

PR #1916, opened 2024-08-07
[JS] feat: ChatCompletion - Tools

corinagum:
## Linked issues

closes: #1576 

## Details
Implement Tools Augmentation which will allow the client to call tools (actions) for OpenAI-like models. Since not all models use the term 'tools' for calling functions, we will continue to use `actions` as a general term. This work does not include adding ToolsAugmentation to other models.

#### Change details

- Implement JS `ToolsAugmentation` - enable this augmentation in `config.json` to facilitate the AI model indicating when to make tool (action) calls
- Bump `openai` to 4.41.1 due to an import error
- Store `actionId` (the equivalent to `tool_call_id` for OpenAI tool calls to `DO` commands
- `OpenAIModel.ts` will convert `actions.json` to tools and send them to the model, which when receiving `tool_calls` from the model will store those as action calls in conversation history in `LLMClient.ts`.
- `ActionPlanner.ts` via `AI.ts` `run` command will call the tools as DO commands and save the output / send back to model in `OpenAIModel.ts`

#### Other:
- Update TeamsAI diagram and add it to `CONCEPTS.md`
- Unit tests coverage increase 

## Attestation Checklist

- [x] My code follows the style guidelines of this project

- I have checked for/fixed spelling, linting, and other errors
- I have commented my code for clarity
- I have made corresponding changes to the documentation (updating the doc strings in the code is sufficient)
- My changes generate no new warnings
- I have added tests that validates my changes, and provides sufficient test coverage. I have tested with:
  - Local testing
  - E2E testing in Teams
- New and existing unit tests pass locally with my changes

PR #1940, opened 2024-08-19
[JS] fix: Streaming Fix

Stevenic:
- we had a bad check for use of the old text completions api.
- added missing logging check.
- added logic to remove tool_choice param when tools not being used.
- fixed a issue related to the turn context being used after the turn completes.

## Linked issues

closes: #1920

## Details

Provide a list of your changes here. If you are fixing a bug, please provide steps to reproduce the bug.

#### Change details

> Describe your changes, with screenshots and code snippets as appropriate

**code snippets**:

**screenshots**:

## Attestation Checklist

  - Local testing
  - E2E testing in Teams

### Additional information


lilyydu:
few more fixes in `StreamingResponse.ts`

1. for `sendTextChunk` rather than sending `text` on sendActivity API call, we need to send `this._message` (message sent is continuously updated per API call)
2. for the "final" stream response, there won't be a `StreamSequence` number in the data body (needs to be an optional field)
3. `this._streamId` has to be set in the `activity.id` field

Stevenic:
> few more fixes in `StreamingResponse.ts`
> 
> 1. for `sendTextChunk` rather than sending `text` on sendActivity API call, we need to send `this._message` (message sent is continuously updated per API call)
> 2. for the "final" stream response, there won't be a `StreamSequence` number in the data body (needs to be an optional field)
> 3. `this._streamId` has to be set in the `activity.id` field

Ok I can add those when I re-work the StreamingResponse class to queue up the chunks internally.

js/packages/teams-ai/src/models/OpenAIModel.ts
```diff
@@ -446,6 +451,13 @@ export class OpenAIModel implements PromptCompletionModel {
         }
         params.model = model;
 
+        // Remove tool params if not using tools
+        if (!Array.isArray(params.tools) || params.tools.length == 0) {
+            if (params.tool_choice) {
+                delete params.tool_choice;
+            }           
+        }
+
```
corinagum, line 460: Not sure this is necessary once Tools Augmentation changes are merged in:
https://github.com/microsoft/teams-ai/pull/1916/files#diff-15e68abdb314274fe327871cd9251953b2946714792b5cd3c401d33b687a9d92R512-R541

Stevenic: It'll still be needed... It's a quirk in their API. They return a 400 error if you include this param (even if set to "auto") and don't provide tools. We're always setting it to "auto" regardless of the augmentation so it has to be removed programmatically

Stevenic:
I updated the PR with requested protocol fixes and I added an outbound activity queue to the `StreamingResponse` class. Incoming chunks from the model are now decoupled from the outgoing chunks to the client. They will be accumulated and sent as fast as they can be sent.  Delivery will always be in sequence order though.

js/packages/teams-ai/src/StreamingResponse.ts
```diff
@@ -53,42 +58,42 @@ export class StreamingResponse {
...
             throw new Error('The stream has already ended.');
         }
 
-        // Send typing activity
-        return this.sendActivity('typing', text, {
-            streamType: 'informative',
-            streamSequence: this._nextSequence++
-        });
+        // Queue a typing activity
+        this.queueActivity(() => ({
```
lilyydu, line 70: wrt the informative update not being sent during yesterday's test session, should we move this into `queueNextChunk `instead

js/packages/teams-ai/src/StreamingResponse.ts
```diff
@@ -100,43 +105,117 @@ export class StreamingResponse {
...
+     */
+    public waitForQueue(): Promise<void> {
+        return this._queueSync || Promise.resolve();
+    }
+
+    /**
+     * Queues the next chunk of text to be sent to the client.
+     * @private 
+     */
+    private queueNextChunk(): void {
```
lilyydu, line 128: each delta that openAI currently returns is a token, could we explore grouping these tokens into a chunk before sending? Confused how we're not hitting a throttle right now but Bot API has a limit of 1 RPS - this might be a problem in the future

corinagum:
@Stevenic FYI - Lily and my tools changes have been checked in. Unfortunately I think there are some conflicts with your changes in this PR. Please let me know if you want to hop on a call to resolve merge conflicts together.

js/packages/teams-ai/src/StreamingResponse.ts
```diff
@@ -100,43 +105,117 @@ export class StreamingResponse {
...
-    private async sendActivity(
-        type: 'typing' | 'message',
-        text: string,
-        channelData: StreamingChannelData
-    ): Promise<void> {
-        // Add stream ID
+    private async sendActivity(activity: Partial<Activity>): Promise<void> {
+        // Set activity ID to the assigned stream ID
         if (this._streamId) {
-            channelData.streamId = this._streamId;
```
lilyydu, line 125: adding back this line

`activity.channelData.streamId = this._streamId;`

lilyydu (approved):

corinagum (approved):

Issue #1968, opened 2024-08-30
[C#] Streaming + Sample


lilyydu:
Enable streaming in C# and update ChefBot sample.

PR #2078, opened 2024-10-02
[C#] feat: Streaming + Sample

lilyydu:
## Linked issues

closes: #1968 

## Details

Added streaming support for C# with an associated sample bot.

#### Change details

- created a separate `IPromptCompletionStreamingModel` (not possible to have optional members for an interface). As a result, OpenAIModel declaration had to be updated.
- converted `StreamingChannelData` into a class

## Attestation Checklist

  - Local testing
  - E2E testing in Teams


dotnet/packages/Microsoft.TeamsAI/Microsoft.TeamsAI/AI/Models/IPromptCompletionModelEvents.cs
```diff
@@ -0,0 +1,170 @@
...
+    /// Events emitted by a IPromptCompletionStreamingModel.
+    /// </summary>
+    public interface IPromptCompletionModelEvents
+    {
+
+        /// <summary>
+        /// Defines the method that is triggered before the model is called to complete a prompt.
+        /// </summary>
+        /// <returns></returns>
+        public delegate void BeforeCompletionHandler(object sender, BeforeCompletionEventArgs args);
```
singhk97, line 19: Is there a reason `BeforeCompletionEventArgs` expects a cancellation token? It should be passed only if the delegate is an "async" delegate ... i.e it returns a `Task` object.

Also the convention is to pass the cancellation token as the last parameter of the delegate. Here's an example: https://github.com/microsoft/teams-ai/blob/e4a556b65892920c171fdbccdb6a8ce7a895c596/dotnet/packages/Microsoft.TeamsAI/Microsoft.TeamsAI/Application/AdaptiveCards/AdaptiveCardsHandlers.cs#L74

singhk97 (commented):
LGTM!

dotnet/packages/Microsoft.TeamsAI/Microsoft.TeamsAI/AI/Models/IPromptCompletionModelEvents.cs
lilyydu: Discussed and removed, forgot to delete

aacebo (commented):
is there a way for us to update the existing chefbot and make streaming configurable? My thought is that instead of having samples for streaming and non streaming, we should treat it like we do with OpenAI/AzureOpenAI, where every samples supports it but its a configuration

aacebo (approved):

PR #2110, opened 2024-10-11
Remove botAadApp/create action in favor of aadApp/create

therealjohn:
botAadApp/create is deprecated in favor of using aadApp/create. This may help also help mitigate issues with Microsoft alias users creating security issues in the MSFT tenant when using samples.


corinagum:
@therealjohn based off our convo a few weeks ago, I'm going to assume this isn't the fix you were looking for previously and will close. Definitely feel free to re-open if that's not the case :)

Issue #1967, opened 2024-08-30
[PY] Streaming + Sample


lilyydu:
Enable streaming in Python and update ListBot sample.

PR #2141, opened 2024-10-25
[PY] feat: Streaming + Sample

lilyydu:
## Linked issues

closes: #1967, 1970

## Details

- Added streaming support for Python with an associated sample bot. 
- TeamsChefBot does not exist, streaming is demonstrated through the ListBot instead.
- Created a custom class to manage events and handlers
- The feedback loop flag is passed in slightly differently from JS and C#. As the AI class was previously not being passed in during `Planner.continue_task `, adding this change will lead to compiling errors for all developers with custom planners. This also led to cyclical dependencies. Instead of internally passing in this flag, it is now exposed as an option on the `ActionPlanner` class. This is more relevant, will not introduce a sudden breaking change to non-streaming developers, and provides more control for streaming developer. Although this allows the flag to deviate from the one in the `AI` class, it is not possible for both flows to run at once.

Additional items for parity:
- Added temporary 1.5 second buffer to adhere to 1RPS backend service requirement
- Added support for Feedback Loop
- Added support for Generated by AI label
- Added reject/catch handling for errors
- Added entities metadata to match GA requirements

**screenshots**:

image

## Attestation Checklist

  - Local testing
  - E2E testing in Teams


python/packages/ai/teams/ai/clients/llm_client.py
```diff
@@ -68,13 +75,29 @@ class LLMClientOptions:
     Optional. When set the model will log requests
     """
 
+    start_streaming_message: Optional[str] = ""
```
singhk97, line 78: if the default is "", should it be an optional type?

singhk97 (approved):

python/packages/ai/teams/ai/clients/llm_client.py
lilyydu: yup! The keyword `Optional` makes it optional

singhk97 (approved):

Issue #2167, opened 2024-11-05
[Feature Request]: Provide two new invoke handlers, (`message/fetchTask`, `message/submitAction`) for Custom Feedback Loops


jk-msft:
## Overview

We currently allow Bot developers to receive feedback from Teams users using [Feedback Loops](https://learn.microsoft.com/en-us/microsoftteams/platform/bots/how-to/bot-messages-ai-generated-content?tabs=before%2Cbotmessage#add-feedback-buttons). When a user clicks "Submit" the feedback, `message/submitAction` invoke is fired. We need a new invoke handler for `message/submitAction` invoke.

We are introducing Bot developers to create **Custom** feedback loops. These custom feedback loops allow Bot developers to show either an Adaptive Card (AC) task module or a website task module. If a bot sent a message with custom feedback loop enabled to the Teams client, the Teams client will fire new invoke, `message/fetchTask` to determine whether the custom feedback loop should show an AC task module or a website task module. We need another invoke handler for `message/fetchTask` invoke.

### Default Feedback

![Image](https://github.com/user-attachments/assets/f0d1d475-b00f-4f69-b1ed-c964325a73f8)

### Custom Feedback (AC)

![Image](https://github.com/user-attachments/assets/340bc4c1-4a6a-4788-8c31-a1a685d69024)

### Custom Feedback (Website)

![Image](https://github.com/user-attachments/assets/ff8d4c68-3bff-442e-87fb-674f9dd426ac)

### Solution

Provide invoke handlers for 
- `message/submitAction`
- `message/fetchTask`

## Return Values

How I am currently handling these two invokes inside `onInvokeActivity`

### `message/submitAction`

```javascript
{
  status: 200,
  body: {}
}
```

#### Adaptive Card

```javascript
const taskModuleReturn = {  
  task: {  
    type: 'continue',  
    value: {  
      card: [Object], // Should contain valid Adaptive Card Payload 
      height: 200,  
      width: 400,  
      title: 'Test Task Module Title with AC'  
    }  
  }  
};
return {
  status: 200,
  body: taskModuleReturn
};
```

#### Website URL

```javascript
const taskModuleReturn = {  
  task: {  
    type: 'continue',  
    value: {  
      url: "https://bing.com", // Should contain valid URL with the valid domain listed under App Manifest
    }  
  }  
};
return {
  status: 200,
  body: taskModuleReturn
};
```

### Send messages with feedback loop

```javascript
    this.onMessage(async (context, next) => {
[...]

PR #2182, opened 2024-11-12
DO NOT MERGE [JS] feat: custom feedback form + citation changes

aacebo:
## Linked issues

closes: #2167 

## Details

update types/handlers for new custom feedback loop and citations changes for ignite.

### DIFF
--- a/js/packages/teams-ai/src/Messages.ts
+++ b/js/packages/teams-ai/src/Messages.ts
@@ -1,0 +1,100 @@
+/**
+ * @module teams-ai
+ */
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import {
+    ActivityTypes,
+    Channels,
+    INVOKE_RESPONSE_KEY,
+    InvokeResponse,
+    TaskModuleResponse,
+    TaskModuleTaskInfo,
+    TurnContext
+} from 'botbuilder';
+import { Application } from './Application';
+import { TurnState } from './TurnState';
+
+export enum MessageInvokeNames {
+    FETCH_INVOKE_NAME = `message/fetchTask`
+}
+
+/**
+ * TaskModules class to enable fluent style registration of handlers related to Task Modules.
+ * @template TState Type of the turn state object being persisted.
+ */
+export class Messages<TState extends TurnState> {
+    private readonly _app: Application<TState>;
+
+    /**
+     * Creates a new instance of the TaskModules class.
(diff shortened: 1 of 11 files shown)


js/packages/teams-ai/src/Messages.ts
```diff
@@ -0,0 +1,100 @@
...
+    ): Application<TState> {
+        this._app.addRoute(
+            async (context) => {
+                return (
+                    context?.activity?.type === ActivityTypes.Invoke &&
+                    context?.activity?.name === MessageInvokeNames.FETCH_INVOKE_NAME
+                );
+            },
+            async (context, state) => {
+                if (context?.activity?.channelId === Channels.Msteams) {
```
jk-msft, line 63: Does this check make it so that only Teams Channels are supported?

Feedbacks will be used in groupchats.

js/packages/teams-ai/src/types/ClientCitation.ts
```diff
@@ -44,22 +68,30 @@ export interface ClientCitation {
         url?: string;
 
         /**
-         * Content of the citation. Must be clipped if longer than 480 characters.
+         * Content of the citation. (max length 160)
```
premkaretimsft, line 71: Let's call this either abstract or excerpt of the citation as it's not the full content. Also, remove the statement about 480 characters (it's an internal SMBA implementation detail) and only say max length is 160.

js/packages/teams-ai/src/types/ClientCitation.ts
```diff
@@ -44,22 +68,30 @@ export interface ClientCitation {
...
+         * Must be clipped if longer than 480 characters.
          */
         abstract: string;
 
         /**
          * Used for icon; for now it is ignored.
          */
-        encodingFormat?: 'text/html';
+        encodingFormat?: 'application/vnd.microsoft.card.adaptive';
```
premkaretimsft, line 79: This value should only be set if `text` is set otherwise, it should be left undefined

js/packages/teams-ai/src/types/ClientCitation.ts
aacebo: the red is the line that was removed, the green is what was added

js/packages/teams-ai/src/Messages.ts
aacebo: this means it will only work in Microsoft Teams

corinagum (dismissed):
LGTM! Thanks Alex!

jk-msft (commented):
Approved.

corinagum (dismissed):

PR #2195, opened 2024-11-20
[JS] feat: streaming support for Tools Augmentation

Stevenic:
## Linked issues

closes: #2197

## Details

This update adds support for properly calling tools with streaming support. It doesn't solve for the other augmentation types as that's much more challenging. 

#### Change details

The challenge with streaming + tool use is that we need to make multiple calls to the model and we can only have a single stream open to the teams client. To address this we end up leveraging the fact that the Streamer is being cached to temp state and we let the tool call happen as it normally world. To the tool call its as if its a non-streaming message but the Streamer object is kept in memory (temp state) waiting for the tool call to complete. We a new request is made to the ActionPlanner, the  created LLMClient will see that it's already created a Streamer and attach to that instance.

There was a bit of book keeping that needed to be done but it's a straightforward change.

**code snippets**:

**screenshots**:

## Attestation Checklist

- [x ] My code follows the style guidelines of this project

  - Local testing
  - E2E testing in Teams

### Additional information


js/packages/teams-ai/src/models/OpenAIModel.ts
```diff
@@ -373,20 +373,44 @@ export class OpenAIModel implements PromptCompletionModel {
                     if (delta.content) {
                         message.content += delta.content;
                     }
+
```
corinagum, line 376: Do these changes need tests?

js/packages/teams-ai/src/models/OpenAIModel.ts
```diff
@@ -373,20 +373,44 @@ export class OpenAIModel implements PromptCompletionModel {
...
-                                    id: toolCall.id,
-                                    function: {
-                                        name: toolCall.function!.name,
-                                        arguments: toolCall.function!.arguments
-                                    },
-                                    type: toolCall.type
-                                } as ActionCall;
+                        // Create action calls array if it doesn't exist
+                        if (!Array.isArray(message.action_calls)) {
+                            message.action_calls = [];
```
lilyydu, line 382: could we add another conditional check that tools augmentation is being used?

`const isToolsAugmentation =
                    template.config.augmentation && template.config.augmentation?.augmentation_type == 'tools';`

js/packages/teams-ai/src/models/OpenAIModel.ts
```diff
@@ -373,20 +373,44 @@ export class OpenAIModel implements PromptCompletionModel {
...
+                        }
+
+                        // Add tool calls to action calls
+                        for (const toolCall of delta.tool_calls) {
+                            // Add empty tool call to message if new index
+                            // - Note that a single tool call can span multiple chunks.
+                            const index = toolCall.index;
```
lilyydu, line 389: just to confirm my understanding, index represents the # of tools that have yet to be completed

js/packages/teams-ai/src/planners/LLMClient.ts
```diff
@@ -347,15 +355,28 @@ export class LLMClient<TContent = any> {
...
-            // - We're not listening for the response received event because we can't await the completion of events.
+            // Handle streaming responses
             if (streamer) {
-                await streamer.endStream();
+                // Tool call handling
+                // - We need to keep the streamer around during tool calls so we're just letting them return as normal
+                //   messages minus the message content. The text content is being streamed to the client in chunks.
+                // - When the tool call completes we'll call back into ActionPlanner and end up re-attaching to the
+                //   streamer. This will result in us continuing to stream the response to the client. 
+                if (Array.isArray(response.message?.action_calls)) {
```
lilyydu, line 366: does this mean we won't be streaming the response until all the tools have been called + returned to the model?

js/packages/teams-ai/src/models/OpenAIModel.ts
Stevenic: Index is the offset in the tool_calls array that they're generating.  Basically they're building up the response they would send you if you weren't streaming so they have to tell you which index they're working on

js/packages/teams-ai/src/planners/LLMClient.ts
```diff
@@ -325,6 +327,12 @@ export class LLMClient<TContent = any> {
                 return;
             }
 
+            // Ignore tool calls
+            // - see the tool call note below to understand why we're ignoring them.
+            if ((chunk.delta as any)?.tool_calls || chunk.delta?.action_calls) {
+                return;
```
lilyydu, line 333: to confirm- are we returning here because the model will only return either streamed chunks **or** tool calls, e.g., never both?

lilyydu: ^ actually this doesn't seem to be the case if we are clearing it in L368- which means that we won't be streaming any text chunks until the tool calls have returned

from a UX standpoint, have you seen this result in a long buffer time with the info message if there are multiple tool calls requested?

getting-started/CONCEPTS/STREAMING.md
```diff
@@ -66,6 +69,7 @@ Once `endStream()` is called, the stream is considered ended and no further upda
 - Attachments can only be sent in the final streamed chunk.
 - Streaming is not available in conjunction with AI SDK's function calls yet.
```
lilyydu, line 70: can remove this line

getting-started/CONCEPTS/STREAMING.md
```diff
@@ -66,6 +69,7 @@ Once `endStream()` is called, the stream is considered ended and no further upda
...
 - Streaming does not work with OpenAI's `o1` models.
+- Tools Streaming only works with the `tools` augmentation. The `sequence` and `monologue` augmentations do not currently support streaming.
```
lilyydu, line 72: can we also note that it works with default augmentation?

getting-started/CONCEPTS/STREAMING.md
```diff
@@ -48,8 +48,6 @@ The expected sequence of calls is:
 2. `queueTextChunk()`, ..., 
 3. `endStream()`.
 
-Once `endStream()` is called, the stream is considered ended and no further updates can be sent.
```
lilyydu, line 48: Sorry should have specified Steve, meant to request removing this line instead
"- Streaming is not available in conjunction with AI SDK's function calls yet."

lilyydu (commented):
code LGTM, thanks Steve! do you mind adding tests (as per Corina's request also)

Stevenic:
I added a not that there isn't a way to test it... the stub class you're using doesn't support stubbing out Server Sent Events.

Issue #1705, opened 2024-06-03
[Bug]: Mising tool outputss Error when using Azure OpenAI Assistants, when the Assistant planner calls the same tool multiple times in a run


NikhilB95:
### Language

Python

### Version

latest

### Description

When using Azure OpenAI Assistants, with custom tools, the teams bot raises a error (shown below) in specific situations when the same tool is called multiple times. 

`Error: Traceback (most recent call last):
  File "/tmp/8dc83d67fdc82b2/antenv/lib/python3.11/site-packages/botbuilder/core/bot_adapter.py", line 174, in run_pipeline
    return await self._middleware.receive_activity_with_status(
  File "/tmp/8dc83d67fdc82b2/antenv/lib/python3.11/site-packages/botbuilder/core/middleware_set.py", line 69, in receive_activity_with_status
    return await self.receive_activity_internal(context, callback)
  File "/tmp/8dc83d67fdc82b2/antenv/lib/python3.11/site-packages/botbuilder/core/middleware_set.py", line 79, in receive_activity_internal
    return await callback(context)
  File "/tmp/8dc83d67fdc82b2/antenv/lib/python3.11/site-packages/teams/app.py", line 663, in on_turn
    await self._start_long_running_call(context, self._on_turn)
  File "/tmp/8dc83d67fdc82b2/antenv/lib/python3.11/site-packages/teams/app.py", line 813, in _start_long_running_call
    return await func(context)
  File "/tmp/8dc83d67fdc82b2/antenv/lib/python3.11/site-packages/teams/app.py", line 756, in _on_turn
    is_ok = await self._ai.run(context, state)
  File "/tmp/8dc83d67fdc82b2/antenv/lib/python3.11/site-packages/teams/ai/ai.py", line 187, in run
    return await self.run(context, state, started_at, step)
  File "/tmp/8dc83d67fdc82b2/antenv/lib/python3.11/site-packages/teams/ai/ai.py", line 187, in run
    return await self.run(context, state, started_at, step)
  File "/tmp/8dc83d67fdc82b2/antenv/lib/python3.11/site-packages/teams/ai/ai.py", line 143, in run
    plan = await self.planner.continue_task(context, state)
  File "/tmp/8dc83d67fdc82b2/antenv/lib/python3.11/site-packages/teams/ai/planners/assistants_planner.py", line 187, in continue_task
    return await self._submit_action_results(state)
[...]

PR #2228, opened 2024-12-10
[PY] fix: #1705 - Output of multiple calls to same tool are overwritten

andres-swax:
The current implementation of the planner did not account for multiple calls to the same tool since it did not keep track of the call id (`call.id`) which is different for each call, regardless of which tool. 

Therefore, the list was only being indexed by `tool_id`.

The result was that subsequent calls to the same tool _within the same run_ would overwrite the result of the previous call, and at the end of the run, results would be missing triggering an Exception.

The issue is fixed by adding a sublevel to the list of tools called as part of a run (this list is stored on assistants_planner.py `tool_map:dict`) which does keep track of the unique call_id, so that their result values are not overwritten by subsequent calls to the same tool within the same run.

## Linked issues

closes: #1705

## Details

In order to trigger this issue, submit a prompt that triggers multiple calls to the same tool on the same run (i.e. which city is warmer today, LA or Chicago?). The same tool would be called once for each city (i.e. get-weather) but the data returned would only contain results for one city. This results on an Exception being raised. Steps can be found on [#1705](https://github.com/microsoft/teams-ai/issues/1705).

#### Change details

> The fix consists on adding a second level to the dictionary of the tool calls (which contains the return values), so instead of being (pseudocode) `tools[tool_id]` it becomes `tools[tool_id][call_id]`. This way the return value of each and every call is retained.

Key change is that `state.temp.action_outputs[command.action_id]` becomes `state.temp.action_outputs[command.action][command.action_id]` on teams/ai/ai.py.

## Attestation Checklist

  - Local testing
  - E2E testing in Teams

### DIFF
--- a/python/packages/ai/teams/ai/ai.py
+++ b/python/packages/ai/teams/ai/ai.py
@@ -171,7 +171,8 @@ class AI(Generic[StateT]):
                     # Set output for action call
                     if command.action_id:
                         loop = True
-                        state.temp.action_outputs[command.action_id] = output or ""
+                        if not command.action in state.temp.action_outputs: state.temp.action_outputs[command.action] = {}
+                        state.temp.action_outputs[command.action][command.action_id] = output or ""
                     else:
                         loop = len(output) > 0
                         state.temp.action_outputs[command.action] = output
--- a/python/packages/ai/teams/ai/planners/assistants_planner.py
+++ b/python/packages/ai/teams/ai/planners/assistants_planner.py
@@ -272,16 +272,17 @@ class AssistantsPlanner(Generic[StateT], _UserAgent, Planner[StateT]):
 
         # Map the action outputs to tool outputs
         action_outputs = state.temp.action_outputs
-        tool_map = state.get(SUBMIT_TOOL_OUTPUTS_MAP)
+        tool_map : dict[str,list] = state.get(SUBMIT_TOOL_OUTPUTS_MAP)
         tool_outputs: List[ToolOutput] = []
 
-        for action in action_outputs:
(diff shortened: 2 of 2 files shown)


andres-swax:

corinagum:
@andres-swax thanks for the contribution! Could you please fix the failing tests and add a new one for the multiple tool calls scenario?

andres-swax:
> @andres-swax thanks for the contribution! Could you please fix the failing tests and add a new one for the multiple tool calls scenario?

Hi, I'll try to do my best, I saw those failures yesterday and honestly, I could not understand a thing nor what I could do.

I felt the failures were on the Python libraries and hoping the messages were just noise.

My dev-ops software and workflow knowledge is still on Visual SourceSafe level or so.  If I cannot make it work, i'll ask for the PR to be closed. On the meantime I'll give it a try for the next cpl of days.

corinagum:
@andres-swax No rush, and we really appreciate your contribution! We may be able to keep the PR open and have someone help out the with tests, but I can't guarantee how soon that will be. Our bandwidth is pretty tight right now. Please add the tests if you can and feel free to ask questions!

andres-swax:
> d we really appreciate your contribution! We may be able to keep the PR open and have someone help out the with tests, but I can't guarant

I tried to see how I can update the tests but tbh I couldn't figure out. If necessary please do close pR so the bug can be taken care of by someone else, it's kind of important to fix it soon.

lilyydu:
Closing the PR - will keep the bug issue open to be picked up by another dev. Thanks @andres-swax for your work!

Issue #2209, opened 2024-12-02
[Bug]: Python SDK introduces double-quotes


jbaron:
### Language

Python

### Version

latest

### Description

When using the Python teams-ai library, the input messages receive additional double-quotes. This doesn't happen with the initial input, but happens with the sub-sequential inputs.

The action is still correct and has just the plain string, but somewhere in the process of converting the action into messages it goes wrong.

So for example when I enter <great> in Teams, I see '"great"' back in the message history (so it has additional double-quotes).

This makes the LLM also use additional quotes in the response or format snippets badly. 

When debugging the JavaScript version, this behavior was NOT observed.

### Reproduction Steps

```bash
1. Enter first prompt (1 + 1) and wait for result
2. Enter second prompt (great)
3. During debugging observe that the message history has additional double-quotes around great.
```

jbaron:
The issue seems to be with the used `to_string` function that doesn't checks for plain strings and as a result uses json or yaml encoding.

```python 

def to_string(tokenizer: Tokenizer, value: Any, as_json: bool = False) -> str:
    """
    Converts a value to a string representation.
    Dates are converted to ISO strings and Objects are converted to JSON or YAML,
    whichever is shorter.

    Args:
        tokenizer (Tokenizer): The tokenizer object used for encoding.
        value (Any): The value to be converted.
        as_json (bool, optional): Flag indicating whether to return the value as JSON string.
        Defaults to False.

    Returns:
        str: The string representation of the value.
    """
    if value is None:
        return ""

    if hasattr(value, "isoformat") and callable(value.isoformat):
        # Used when the value is a datetime object
        return value.isoformat()
    value = todict(value)

    if as_json:
        return json.dumps(value, default=lambda o: o.__dict__, ensure_ascii=False)

    # Return shorter version of object
    yaml_str = yaml.dump(value, allow_unicode=True)
    json_str = json.dumps(value, default=lambda o: o.__dict__, ensure_ascii=False)
    if len(tokenizer.encode(yaml_str)) < len(tokenizer.encode(json_str)):
        return yaml_str

    return json_str

```

If the value is of the type string, it shouldn't use JSON or YAML encoding.

Nivedipa-MSFT:
@jbaron - Thank you for your inquiry about your Teams app development issue! We will check and update you soon.

sayali-MSFT:
Hello @jbaron ,Thank you for your patience! We have reported this as a bug for further investigation. 

[...]

PR #2259, opened 2025-01-10
[PY] fix: Don't use json/yaml formatting for simple string values

jbaron:
## Linked issues

closes: #2209 

## Details

Changed the Python `to_string` function to not convert the value object if it is just a plain string. This caused the introduction of doubled quotes and made responses render badly in MS Teams.

#### Change details

> Just added a single check to the to_string function that in case of a string, don't do anything and just return the value.

## Attestation Checklist

### Additional information


corinagum:
Hi @jbaron thanks for the submission! It looks like there are several test failures - could you please run the tests locally and take a look? I think your solution may be having a wider area of effect than anticipated. 

here's an example:

```
    async def test_render_as_messages(self):
        function_response_message = FunctionResponseMessage("foo", "bar")
        rendered = await function_response_message.render_as_messages(
            self.turn_context, self.memory, self.prompt_functions, GPTTokenizer(), 100
        )
        self.assertEqual(rendered.output[0].role, "function")
>       self.assertEqual(rendered.output[0].content, '"bar"')
E       AssertionError: 'bar' != '"bar"'
E       - bar
E       + "bar"
```

jbaron:
I updated the test cases and `poetry run test` now is successful. But:

- On my Mac I had to update the tiktoken to 0.8.0, otherwise I could not install it without compiling from source. 
- Only Python 3.9 and higher is supported by this tiktoken version, so also changed this

So not sure this additional changes are desirable.

jbaron:
I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.

jbaron:

jbaron:
> > I updated the test cases and `poetry run test` now is successful. But:
> > 
> > * On my Mac I had to update the tiktoken to 0.8.0, otherwise I could not install it without compiling from source.
> > * Only Python 3.9 and higher is supported by this tiktoken version, so also changed this
> > 
> > So not sure this additional changes are desirable.
> 
> Hi @jbaron, what error were you getting with tiktoken that caused you to update?

The problem was (if I recall correctly) that there was no binary wheel for tiktoken for my MacOS (15.2). So the package installer tried to compile it from the source and failed because it couldn't find a Rust compiler. 

A later version of tiktoken had a binary wheel for my OS, but didn't support the older versions of Python.

lilyydu:
> > > I updated the test cases and `poetry run test` now is successful. But:
> > > 
> > > * On my Mac I had to update the tiktoken to 0.8.0, otherwise I could not install it without compiling from source.
> > > * Only Python 3.9 and higher is supported by this tiktoken version, so also changed this
> > > 
> > > So not sure this additional changes are desirable.
> > 
> > 
> > Hi @jbaron, what error were you getting with tiktoken that caused you to update?
> 
> The problem was (if I recall correctly) that there was no binary wheel for tiktoken for my MacOS (15.2). So the package installer tried to compile it from the source and failed because it couldn't find a Rust compiler.
> 
> A later version of tiktoken had a binary wheel for my OS, but didn't support the older versions of Python.

Gotcha, we are erring on preventing the bump for Python. Since this seems to be specific to your device, would it be possible for you to try installing Rust and see if that fixes it? (linking this [guide](https://stackoverflow.com/questions/75124685/unable-to-install-tiktoken-from-pypi))

lilyydu:
> > > > > I updated the test cases and `poetry run test` now is successful. But:
> > > > > 
> > > > > * On my Mac I had to update the tiktoken to 0.8.0, otherwise I could not install it without compiling from source.
> > > > > * Only Python 3.9 and higher is supported by this tiktoken version, so also changed this
> > > > > 
> > > > > So not sure this additional changes are desirable.
> > > > 
> > > > 
> > > > Hi @jbaron, what error were you getting with tiktoken that caused you to update?
> > > 
> > > 
> > > The problem was (if I recall correctly) that there was no binary wheel for tiktoken for my MacOS (15.2). So the package installer tried to compile it from the source and failed because it couldn't find a Rust compiler.
> > > A later version of tiktoken had a binary wheel for my OS, but didn't support the older versions of Python.
> > 
> > 
> > Gotcha, we are erring on preventing the bump for Python. Since this seems to be specific to your device, would it be possible for you to try installing Rust and see if that fixes it? (linking this [guide](https://stackoverflow.com/questions/75124685/unable-to-install-tiktoken-from-pypi))
> 
> I think this is an issue with all Macs that have the latest OS installed, but this is of course only an issue for people that want to develop and test the team-ai library itself (not too many I guess). If you only want to use teams-ai, no problem.
> 
> On this machine I cannot install Rust (policy), but if I have some time I can try it on another (Linux) machine with the old tiktoken version if that helps (or perhaps just a GitHub action). But that will take some time.

Thank you, that would be great! In the meantime, will check with the mac users on our team to see if this is an issue.

lilyydu (approved):

PR #2379, opened 2025-04-04
[C#] fix: Add codesigning to .NET build pipeline

singhk97:
## Linked issues

closes: #minor

## Details

Bump to `v1.10.1`

Update dotnet release pipeline to do the following:

1. Strong name sign .dll
image
2. Authenticode sign .dll
image
3. Sign nuget package
image

## Attestation Checklist

  - Local testing
  - E2E testing in Teams


pipelines/dotnet.yml
```diff
@@ -41,35 +135,44 @@ steps:
...
+    nobuild: true
+    configuration: '$(buildConfiguration)'
+
+- task: EsrpCodeSigning@5
+  displayName: 'Sign Nuget package'
   inputs:
-    ConnectedServiceName: 'Teams ESRP CodeSign v3'
+    ConnectedServiceName: 'TeamsESRP-CP-401405'
+    AppRegistrationTenantId: cdc5aeea-15c5-4db6-b079-fcadd2505dc2
+    AppRegistrationClientId: a94a770a-9a7b-4888-a3ea-24584b851e49
```
lilyydu, line 146: double checking- are these all okay as public vals?

pipelines/dotnet.yml
```diff
@@ -27,12 +28,105 @@ steps:
...
     projects: '$(package)'
 
+- powershell: |
+    # This allows us to not have to checkin .csproj files with DelaySign and SignAssembly set to to true. If the flag is set,
+    # then project is not debuggable with SignAssembly set to true.
+    $csprojPath = @(".\dotnet\packages\Microsoft.TeamsAI\Microsoft.TeamsAI\Microsoft.Teams.AI.csproj")
+    $doc = New-Object System.Xml.XmlDocument
+    $doc.Load($csprojPath)
+    # Set the DelaySign element to 'true' so that delay signing is set.
+    $delaySign = $doc.SelectSingleNode("//DelaySign");
```
KirillOsenkov, line 38: You should use PublicSign instead of DelaySign. 

Read about PublicSign here:
https://github.com/dotnet/runtime/blob/main/docs/project/public-signing.md

KirillOsenkov:
This strong name signing approach is objectively worse than what was there before.

It doesn't allow third parties to strong name sign themselves, and have strong name signing work during debugging. The guidance for open source projects is to not use the Microsoft strong name key, but instead check in a randomly generated key, so that third parties are on the same level playing field as the first party.

Another severe downside is that it mutates the files. You have non-strong-named files, and then they have to mutate during the signing process. The whole build process becomes more complicated, brittle and error prone. However unfortunately you have to do this for Authenticode anyway, so this is less of an issue.

I'm not sure why you chose to go this way for strong naming, but it is ultimately your team's choice.

pipelines/dotnet.yml
singhk97: https://github.com/microsoft/OpenAPI.NET/blob/25136afa5a35160c75ab6e8bd297fbab479efd74/.azure-pipelines/ci-build.yml#L85

I found a handful of OSS repos that have these ids publically available. I think it should be fine.

heyitsaamir (dismissed):

lilyydu (dismissed):

aacebo (approved):

PR #2394, opened 2025-04-21
Restructure docs

heyitsaamir:
Mainly moving some files around to be closer to the structure we had all discussed.

image


book/src/SUMMARY.md
```diff
@@ -1,51 +1,56 @@
 # Summary
 
 - [👋 Welcome](./1.welcome/README.md)
+  - [Why an SDK?](./1.welcome/1.need-for-sdk.md)
```
corinagum, line 4: This is a nit and doesn't need to be addressed in this PR: 

I think it would be good to be more specific and say "Why Teams AI v2?" here. It gives the SDK's name more visibility and I assume we'll be more focused on justifying the "v2" part rather than "SDK" part, if that makes sense.

corinagum (commented):
LGTM! Thanks for getting these changes started @heyitsaamir

book/src/SUMMARY.md
heyitsaamir: Actually this doc in particular currently deals with why do we need an SDK at all. To alex's point, it's more lower level. I think why teams ai sdk v2 should be an entirely new section!

PR #2396, opened 2025-04-22
v2 - updated welcome section

lilyydu:


book/src/1.welcome/2.package-decomposition.md
```diff
@@ -0,0 +1,53 @@
+# Package Decomposition
+The SDK is broken down into the following packages.
```
heyitsaamir, line 2: Keep in mind this will change when C# comes in (very soon). Any ideas on how this might evolve?

book/src/1.welcome/README.md
```diff
@@ -1,54 +1,19 @@
-# 📖 Teams SDK: Typescript: Documentation
+# 📖 Teams SDK v2
 
-Microsoft Teams is a feature rich messaging platform. It provides features that developers can use
-to create immersive app experiences, below are some of the different packages/features that can be used with this SDK.
+Teams SDK v2 is a suite of packages used to build on Microsoft Teams Platform.
```
heyitsaamir, line 3: Wdyt of something along the lines of how its been rebuilt from the ground up? Idk if that's controversial.

heyitsaamir, line 3: to build what on teams?

book/src/1.welcome/README.md
```diff
@@ -1,54 +1,19 @@
...
 
-## [`@microsoft/teams.apps`](https://www.npmjs.com/package/@microsoft/teams.apps)
+## ⭐ What's new?
 
-Tools for building server side apps for Microsoft Teams.
+### Streamlined Developer Experience
+We’ve refined the developer experience so you can concentrate on building your app’s logic— not wrestling with integrations.
```
heyitsaamir, line 8: some trivial example?

heyitsaamir: Also something about devtools?

book/src/1.welcome/README.md
```diff
@@ -1,54 +1,19 @@
...
+### Effortless Integration
+We’ve simplified complex integration workflows to help you deliver a richer, more seamless user experience.
 
-Core types and client implementations for the Teams Platform API,
-used for fetching data, authentication, and sending activities.
+### Accelerate Your Workflow
+Get your application up and running in under 30 seconds with our lightning-fast CLI—so you can spend more time on what really matters.
 
-## [`@microsoft/teams.ai`](https://www.npmjs.com/package/@microsoft/teams.ai)
+## 🔎 Navigation Tips
```
heyitsaamir, line 16: Should we include CTA? Where to file issues, ask questions etc? Maybe we need to figure that out first...

book/src/1.welcome/2.package-decomposition.md
lilyydu: are the learn docs our source for language-specific questions? if thats the case, maybe we can have this included there instead

book/src/1.welcome/README.md
lilyydu: powerful agents? that sorta seems repetitive in the industry tho

corinagum: 'rebuilt from the ground up with ease of use in mind' or something like that.

book/src/1.welcome/README.md
corinagum: I would expect to find 'github related' info on the readme of the repo, not inside these docs. Previously I had Contributing docs in mdbook, but Alex pointed out it makes more sense to live on the repo. I think the same makes sense for issues.

For filing issues/asking questions, I think we decided we DON'T want to keep the Discussions section open going forward, so 'dev questions' would either be a type of issue to file and/or link to the Teams channel for support. 

My 2 cents.

----

Do we want something like MSDocs has where users can file an issue directly to one of our docs files if they notice a typo or discrepancy?

book/src/1.welcome/2.package-decomposition.md
corinagum: I believe we will be adding the Tabs plugin to mdbook which would allow us to put different lists of the packages per language like

TypeScript | C# | Python

book/src/1.welcome/2.package-decomposition.md
```diff
@@ -0,0 +1,53 @@
...
+
+A client used to create app/bot surfaces such as tabs using `@microsoft/teams.ts`.
+
+## [`@microsoft/teams.common`](https://www.npmjs.com/package/@microsoft/teams.common)
+
+Common implementations used by all the packages, for example `logging`.
+
+## [`@microsoft/teams.dev`](https://www.npmjs.com/package/@microsoft/teams.dev)
+
+Developer tools to streamline the development process.
```
corinagum, line 44: I would add something like

book/src/1.welcome/README.md
corinagum: '..build agents and apps that take advantage of Teams features like message extensions, ....'

book/src/1.welcome/README.md
corinagum, line 8:
```suggestion
We’ve refined the developer experience so you can concentrate on building your app’s logic — not wrestling with integrations.
```

book/src/1.welcome/2.package-decomposition.md
lilyydu: hmm in that case, should we delete this for now until we introduce tabs?

book/src/1.welcome/README.md
lilyydu: I agree with keeping it on the Github repo, direct filings would be nice as well

PR #2425, opened 2025-05-02
Remove numbering in filenames

heyitsaamir:
Removes numbering from filenames in docs as discussed. This is so that our links don't have numbering in them, and so that we don't have to manage ordering of file-names

Also fixes a bunch of broken links. While doing this, I ended up creating a script that checked for validity of links. I'm adding that as a github workflow actino.


book/src/SUMMARY.md
```diff
@@ -1,61 +1,61 @@
...
+  - [App Basics](./essentials/app-class.md)
+  - [Listening To Events](./essentials/on-event.md)
+  - [Listening To Activities](./essentials/on-activity.md)
+    - [Activity Reference](./essentials/activity-ref.md)
+  - [Sending Messages](./essentials/sending-messages.md)
+    - [Proactive Messages](./essentials/proactive-messaging.md)
+  - [Using The API Client](./essentials/api.md)
+- [Teams](./teams/README.md)
+  - [Core Concepts](./teams/core-concepts.md)
+  - [Teams Manifest](./teams/manifest.md)
```
heyitsaamir, line 20: @corinagum ok to put in?

corinagum: Thanks for the catch - forgot to do it :/

scripts/check-links.js
```diff
@@ -0,0 +1,330 @@
+#!/usr/bin/env node
+
+const fs = require('fs');
```
corinagum, line 3: I think I added these script files to .prettierignore (before @aacebo made the change to eslint) - could you validate whether the scripts in this folder still need to be added or can be removed? At the least I think the .min.js file should remain ignored.

heyitsaamir: Looks like these scripts aren't ignored. That's probably ok tbh

Issue #2480, opened 2025-05-22
[Feature Request]: Document how to access graph endpoints currently not exposed


mkesper:
### Scenario

Currently only a part of graph API endpoints is exposed as clients.
How to access the other ones?
Add an example showing how to access the raw API.
There probably will always be new (beta) endpoints not yet exposed that someone needs to access from Teams.

### Solution

Add an example to graph documentation (graph.md) like drafted in https://github.com/microsoft/teams-ai/pull/2479

corinagum:
Hi @mkesper, thanks for the feedback! 
For the endpoints we don't support, feel free to peruse the [Learn endpoint reference docs](https://learn.microsoft.com/en-us/graph/api/overview?view=graph-rest-1.0&preserve-view=true)

We don't want to be misleading by documenting the endpoints we don't support, but we can definitely point to where to find that information in our docs. Keeping the docs updated to list the endpoints we DO support is very important though.

@Jesperholmbergmsft is currently working on adding more content to the graph related docs, including keeping the supported endpoints updated. Please keep an eye out for that :)

mkesper:
Thanks @corinagum for the answer. 
@Jesperholmbergmsft The tricky part for me was getting the user token as there seems no (obvious) way to get it in an on.message handler, only on a signin event handler. 
If you could add some hints regarding that it would probably be helpful for others.

Jesperholmbergmsft:
@mkesper  - thanks for raising this.

As a first step towards improving the situation, I opened this PR to fully document exactly which endpoints we do support already: https://github.com/microsoft/teams-ai/pull/2498. This should at least help save time searching for things that aren't available.

For the specific question, the graph client doesn't currently make it easy to call endpoints that we don't yet support. I'd like to improve the graph client and build out the support further, so you get typing support and all. Ideally the client would be at a state where it supports everything you need, so you don't need to construct the requests by hand. But if I can't achieve that, I'll keep your request in mind to see how best to provide an escape hatch.

Jesperholmbergmsft:
@mkesper  - we've redesigned the Graph implementation this summer, and this allowed us to grow the supported APIs tremendously. Should you see any gaps, there's now an affordance that lets you specify the API path, HTTP method, parameters etc.

Please see https://microsoft.github.io/teams-ai/typescript/essentials/graph and https://microsoft.github.io/teams-ai/typescript/migrations/preview for details.

As for doing this in a message event handler, there's a userGraph prop in the context that can be used:
```
[...]

PR #2498, opened 2025-06-06
Improve graph client documentation

Jesperholmbergmsft:
## Linked issues

Related to: https://github.com/microsoft/teams-ai/issues/2480

## Details

The Graph API client does not support the full and entire Microsoft Graph API surface. It's difficult for developers to understand which endpoints are actually supported and how to access them. 

As a first step towards improving the graph client, this PR adds in-depth guides to show exactly which endpoints are currently supported. 

#### Change details
The change here adds a new in-depth overview page for the graph client, and subpages for each area that is supported. The main audience for this change is someone who's already familiar with the Graph API surface and needs to understand what this library provides and how to access those features. 

The documentation stops short at showing input parameter types and return values. We should add that too in the future, but this is a good start.

**screenshots**:
New readme.md start page:
image

Child page:
image

## Attestation Checklist

- [ x ] My code follows the style guidelines of this project
- [ x ] I have checked for/fixed spelling, linting, and other errors
- [ x ] My changes generate no new warnings

### DIFF
--- a/teams.md/docs/typescript/in-depth-guides/graph/app-catalogs.md
+++ b/teams.md/docs/typescript/in-depth-guides/graph/app-catalogs.md
@@ -1,0 +1,65 @@
+# App Catalogs
+
+This page lists all the `/appCatalogs` graph endpoints that are currently supported in the Graph API Client. The supported endpoints are made available as type-safe and convenient methods following a consistent pattern.
+
+You can find the full documentation for the `/appCatalogs` endpoints in the [Microsoft Graph documentation](https://learn.microsoft.com/en-us/graph/api/appcatalogs-list-teamsapps?view=graph-rest-1.0), including details on input arguments, return data, required permissions, and endpoint availability.
+
+## Getting Started
+
+To get started with the Graph Client, please refer to the [Graph API Client Essentials](../../essentials/graph) page.
+
+
+## AppCatalogsClient Endpoints
+
+The AppCatalogsClient instance gives access to the following `/appCatalogs` endpoints. You can get a `AppCatalogsClient` instance like so:
+
+```typescript
+const appCatalogsClient = graphClient.appCatalogs;
+```
+| Description | Endpoint | Usage | 
+|--|--|--|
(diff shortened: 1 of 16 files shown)


teams.md/docs/typescript/in-depth-guides/graph/app-catalogs.md
```diff
@@ -0,0 +1,65 @@
...
+## TeamsAppsClient Endpoints
+
+The TeamsAppsClient instance gives access to the following `/appCatalogs/teamsApps` endpoints. You can get a `TeamsAppsClient` instance like so:
+
+```typescript
+const teamsAppsClient = appCatalogsClient.teamsApps;
+```
+| Description | Endpoint | Usage | 
+|--|--|--|
+| List teamsApp | `GET /appCatalogs/teamsApps` | `await teamsAppsClient.list(params);` |
```
heyitsaamir, line 33: instead of "teamsAppsClient", should we use "graphClient"? And clarify that "graphClient" could come from a number of different places, and the main differences are the token that's supplied to it?

heyitsaamir (commented):
Looks great. One minor nit question

teams.md/docs/typescript/in-depth-guides/graph/app-catalogs.md
Jesperholmbergmsft: @heyitsaamir  - `teamsAppsClient` here is the class that implements the `/appCatalogs/teamsApps`/ endpoints. We can probably improve the naming of these things, but for the moment they're at least consistent for all endpoints.

The developer gets the `teamsAppsClient` instance from `graphClient.appCatalogs.teamsApps`, but I didn't want to put that in the table here. It gets too redundant and too verbose, the table becomes unreadable. I'm hoping that the note before the table and the Getting Started section at the top of each page would be sufficient for the reader to learn how to find a graph client, and then how to find these implementations too. What do you think? Would it be helpful to repeat the Getting Started link for each client type..?

PR #2604, opened 2025-09-15
fix: lang to teams new command docs

rido-min:
In UserAuth quickstarts, the `teams new` command is missing the first param with the language


AfreenLikeCaffeine (approved):

heyitsaamir (approved):
We had introduced a "set-lang" feature too, but this is more correct. Mind adding a :::tip to indicate that they can "set-lang" to csharp if they don't want to always include it as part of their cli commands?

PR #2634, opened 2025-09-25
[docs]: added v1 to v2 migration guide for python

lilyydu:
adds migration section and general guide from v1 to v2 for python
- "action planner" section has to be added once 'agent' implementation is updated

### DIFF
--- a/scripts/generate-snippets.js
+++ b/scripts/generate-snippets.js
@@ -65,7 +65,7 @@ const getDefaultConfig = (lang) => {
       };
     case 'dotnet':
       return {
-        input: path.join(projectRoot, 'teams.dotnet'), // Assuming dotnet input path
+        input: path.join(projectRoot, 'teams.dotnet/Samples'),
         '--output': [path.join(projectRoot, 'teams.md/static/generated-snippets/dotnet')],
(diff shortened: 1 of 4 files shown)


teams.md/docs/python/migrations/v1/README.md
```diff
@@ -0,0 +1,507 @@
...
+from microsoft.teams.apps import ActivityContext, App, ErrorEvent
+from microsoft.teams.common import LocalStorage
+
+logger = logging.getLogger(__name__)
+
+# Define the app
+app = App()
+
+# Optionally create local storage
+storage: LocalStorage[str] = LocalStorage()
```
heyitsaamir, line 120: doesn't this need to go in the App? also do we need this?

heyitsaamir (approved):

teams.md/docs/python/migrations/v1/README.md
lilyydu: nope, I think it's just to demonstrate having local storage outside of the app, was copying TS!
https://microsoft.github.io/teams-ai/typescript/migrations/v1/#migrate-application-class

PR #2724, opened 2026-03-10
[docs] Adaptive Cards: Add note on AC version

corinagum:
- Add note on Adaptive Cards version being required for proper rendering on mobile


teams.md/src/pages/templates/in-depth-guides/adaptive-cards/building-adaptive-cards.mdx
```diff
@@ -32,6 +32,10 @@ Benefits:
 
 <LanguageInclude section="source-code-note" />
 
+:::note
+`AdaptiveCard` defaults the `version` property to `"1.5"`. The `version` field is required for Adaptive Cards to render on Teams mobile clients.
```
heyitsaamir: Hm, is this worth including? We should do the right thing for them right?

PR #2769, opened 2026-04-20
Docs: expand logging guide with levels, patterns, env vars; fix broken Python example

corinagum:
## Summary

Expands `in-depth-guides/observability/logging` to cover material that was previously missing, and fixes a broken Python example that referenced SDK symbols that don't exist.

**What's added (TS & PY):**
- **Log Levels** — full list of available levels and default behavior.
- **Filtering by Logger Name** — `pattern` option in code (TS), `ConsoleFilter` in code (PY), wildcard and exclusion syntax.
- **Environment Variables** — `LOG_LEVEL` and `LOG` for TS (with a callout that `LOG` must match the default logger name `@teams/app`, or it will silence output). Note for PY that env vars are not read automatically by the SDK.
- **Child Loggers** — `log.child('scope')` pattern (TypeScript only; N/A for Python).

**Python fix:**
The previous `python.incl.md` example imported `ConsoleLogger` and `ConsoleLoggerOptions` from `microsoft_teams.common` and called `App(logger=...)` — none of which exist. Replaced with the canonical pattern from `teams.py/examples/oauth/src/main.py`: stdlib `logging` + `ConsoleFormatter`. A user copy-pasting the old example would have hit `ImportError`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/teams.md/src/components/include/in-depth-guides/observability/logging/csharp.incl.md
+++ b/teams.md/src/components/include/in-depth-guides/observability/logging/csharp.incl.md
@@ -23,3 +23,19 @@ builder.AddTeams(appBuilder)
 var app = builder.Build();
 var teams = app.UseTeams();
 ```
+
+<!-- log-levels -->
+
+N/A
+
+<!-- pattern-example -->
+
+N/A
+
+<!-- env-vars -->
+
+N/A
+
+<!-- child-logger -->
+
+N/A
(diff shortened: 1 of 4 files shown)


lilyydu (approved):

teams.md/src/components/include/in-depth-guides/observability/logging/csharp.incl.md
```diff
@@ -23,3 +23,19 @@ builder.AddTeams(appBuilder)
 var app = builder.Build();
 var teams = app.UseTeams();
 ```
+
+<!-- log-levels -->
+
+N/A
```
rido-min, line 29: @corinagum  dont we support logging config via envvars in C#?

PR #2773, opened 2026-04-22
cli: validate MSAL and Azure CLI tenants match

heyitsaamir:
## Summary
- adds `ensureTenantMatch()` utility that throws `TENANT_MISMATCH` if the Teams login tenant differs from the Azure CLI tenant
- `app create` and `bot migrate` now block on tenant mismatch before creating any resources
- `status` command shows tenant match/mismatch with both IDs and a fix suggestion
- adds unit tests for the new tenant check functions

## Test plan
- run `teams status` while logged into different tenants in Teams vs Azure CLI — should show red "mismatch"
- run `teams status` with matching tenants — should show green "matches Teams login"
- run `teams app create --azure` with mismatched tenants — should throw TENANT_MISMATCH error
- `npx vitest run` passes


singhk97 (approved):

singhk97 (approved):

rido-min:
What's the downside of supporting both? Eg, my M365 Developer tenant does not have sub, so I'm using my MSDN sub, and this results in two different tenants

PR #2789, opened 2026-04-25
cli: validate endpoint and inputs before creating resources

heyitsaamir:
## Summary
- Adds `validateEndpoint()` to catch bad URLs (placeholders, HTTP, malformed) before any resources get created
- Moves all field validations (name length, description length, URL format) upfront in `app update` so a bad flag doesn't leave the app half-updated
- Adds inline `validate` on interactive endpoint prompts in both `create` and `update` so users get immediate feedback

## Test plan
- [x] `teams app create --name Test --endpoint "https://<foo>/api" --yes` → fails immediately with "Endpoint is not a valid URL."
- [x] `teams app create --name Test --endpoint "http://example.com/api" --yes` → fails with "Endpoint must use HTTPS."
- [x] `teams app update <id> --endpoint "ftp://bad" --json` → fails with validation error before any API calls
- [x] `teams app update <id> --name "x]x31" --json` → fails upfront, no partial mutations
- [x] Interactive `app create` → entering a bad endpoint re-prompts instead of accepting
- [x] `npm run test` passes

### DIFF
--- a/packages/cli/src/apps/manifest.ts
+++ b/packages/cli/src/apps/manifest.ts
@@ -50,10 +50,6 @@ export function validateEndpoint(url: string): string | null {
   if (parsed.hostname === 'localhost' || parsed.hostname === '127.0.0.1') {
     return null; // allow localhost for dev
   }
-  // Reject placeholder/template URLs like <your-tunnel-url>
-  if (parsed.hostname.includes('<') || parsed.hostname.includes('>')) {
-    return 'Endpoint contains placeholder characters — replace with an actual URL.';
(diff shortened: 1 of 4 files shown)


heyitsaamir:
Re: httpsUrlRegex not enforcing hostname — good catch in principle, but `https:///path` already throws in Node's URL parser so the regex can't actually accept it. The regex is pre-existing code not introduced by this PR — happy to tighten it in a follow-up but keeping out of scope here.

singhk97 (approved):

PR #2833, opened 2026-05-12
fix(cli): surface real cause when discoverAzureBot fails

### LINKED ISSUE #2830: [Bug]: discoverAzureBot swallows all errors as "not found", causing misleading messages and risk of duplicate bot creation
### Bug Description

`discoverAzureBot` in `packages/cli/src/apps/bot-handler.ts` wraps its lookup in a bare `catch { return null }` (lines 225–228). This collapses every distinguishable failure mode: `az` not installed, `az` not logged in, network error, malformed JSON, permission denied, ARM 5xx all into the same `null` return value that signals "this bot legitimately doesn't exist in Azure" (line 212).

```ts
// packages/cli/src/apps/bot-handler.ts:198–229
export async function discoverAzureBot(
  botId: string,
  silent = false
): Promise<AzureContext | null> {
  const spinner = createSilentSpinner('Discovering Azure bot...', silent).start();
  try {
    const results = await runAz<Array<{ resourceGroup: string; location: string }>>([
      'resource', 'list', '--resource-type', 'Microsoft.BotService/botServices', '--name', botId,
    ]);
    if (results.length === 0) {
      spinner.stop();
      return null;                                // legitimate "not found"
    }
    const bot = results[0];
    const account = await runAz<{ id: string; tenantId: string }>(['account', 'show']);
    spinner.success({ text: 'Azure bot discovered' });
    return { ... };
  } catch {
    spinner.stop();
    return null;                                  // swallows every error as "not found"
  }
}
```

Callers then react incorrectly:

- `commands/app/update.ts:273–276` (interactive endpoint update): displays *"Could not find this bot in Azure"* for every underlying cause, which is factually wrong for most of them.
- `commands/app/update.ts:511–518` (non-interactive, `--yes` / `--json`): throws `CliError('NOT_FOUND_AZURE_BOT', ...)` even when the real error is auth, network, or permissions. JSON consumers receive a misleading error code, and an automated agent running with `--yes` is at risk of attempting to recreate a bot that already exists but couldn't be read.
- `commands/app/doctor.ts:220` is the only caller that pre-checks `isAzInstalled()` / `isAzLoggedIn()` before calling, so it incidentally avoids the worst of the smell, but that defensive pattern isn't repeated elsewhere.

The existing `ErrorCode` taxonomy in `src/utils/errors.ts` already defines `TOOL_AZ_NOT_INSTALLED`, `TOOL_AZ_NOT_LOGGED_IN`, `NOT_FOUND_AZURE_BOT`, `PERMISSION_AZURE_REQUIRED`, and `API_ARM_ERROR`. All the discriminators a correct implementation would need are already available; `discoverAzureBot` simply discards them.

### Steps to Reproduce

Minimal standalone repro that simulates `az` being unavailable (a stand-in for any underlying `runAz` failure):

1. From the monorepo root: `npm install --ignore-scripts`
2. `cd packages/cli && npx tsc`
3. Create `packages/cli/repro-f2.mjs`:
   ```js
   process.env.PATH = '';
   process.env.PATHEXT = '';
   const { discoverAzureBot } = await import('./dist/apps/bot-handler.js');
   const result = await discoverAzureBot('fake-bot-id-for-repro', true);
   console.log('Returned:', result);
   ```
4. `node repro-f2.mjs`

Output:

```
Returned: null
```

The function returns `null` despite `az` being completely unreachable. This is the same value it would return if the bot legitimately didn't exist. The caller cannot distinguish "I asked and the answer is no" from "I couldn't ask at all."

### Expected Behavior

`discoverAzureBot` reserves `null` for the legitimate zero-results case already handled at `bot-handler.ts:212`. All other failure modes are rethrown as `CliError` with an appropriate `ErrorCode` (e.g. `TOOL_AZ_NOT_LOGGED_IN`, `PERMISSION_AZURE_REQUIRED`, `API_ARM_ERROR`), so callers and `--json` consumers see the real cause.

### Actual Behavior

Every failure inside the `try` block becomes `null`. Callers report *"Could not find this bot in Azure"* (or `{"ok": false, "error": {"code": "NOT_FOUND_AZURE_BOT", ...}}` in `--json` mode) regardless of the real underlying cause.

### SDK Version

`@microsoft/teams.cli` at `main` (HEAD `f4b2fb17`)

### Languages Affected

TypeScript

### Additional Context

Repro confirmed on Windows 11, Node 20, against current `main`.

singhk97:
Fixes #2830.

`discoverAzureBot` and its helpers collapsed every az failure into `null`, indistinguishable from a real "bot not found." `app update --json` then returned a misleading `NOT_FOUND_AZURE_BOT`; under `--yes`, an agent could try to recreate a bot that already exists but couldn't be read.

Rebased on top of #2825's fallback chain. Only the bug-shaped catches are removed; the two structural ones are kept:

- `findBotByName` outer catch — **removed** (pure suppression).
- `findBotByMsaAppIdViaList` outer catch — **removed** (pure suppression).
- `findBotByMsaAppIdViaList` per-bot catch — **kept** (per-element resilience inside `Promise.all`).
- `findBotByMsaAppIdViaGraph` outer catch — **kept** (intentional fallback signal: routes "graph extension unavailable" to the list path; pinned by #2825's tests).

`discoverAzureBot` rewraps the surviving catch as `CliError('API_ARM_ERROR', 'Failed to discover Azure bot: <stderr>')` — preferring `error.stderr` over `error.message` so the real az diagnostic surfaces instead of `Command failed: cmd /c az.cmd ...`. `app doctor` wraps the call so a throw becomes a `fail` entry instead of terminating the report.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/cli/src/commands/app/doctor.ts
+++ b/packages/cli/src/commands/app/doctor.ts
@@ -217,7 +217,18 @@ async function checkBotRegistration(
         detail: 'Run az login',
       });
     } else {
-      azure = await discoverAzureBot(botId, silent);
+      let discoveryFailed = false;
+      try {
+        azure = await discoverAzureBot(botId, silent);
+      } catch (e) {
+        discoveryFailed = true;
+        results.push({
+          category: cat,
+          label: 'Could not discover Azure bot',
+          status: 'fail',
+          detail: e instanceof Error ? e.message : undefined,
+        });
+      }
       if (azure) {
         results.push({
           category: cat,
@@ -288,7 +299,7 @@ async function checkBotRegistration(
             detail: e instanceof Error ? e.message : undefined,
           });
         }
-      } else {
+      } else if (!discoveryFailed) {
         results.push({
           category: cat,
           label: 'Azure bot not found',
--- a/packages/cli/src/apps/bot-handler.ts
+++ b/packages/cli/src/apps/bot-handler.ts
@@ -4,6 +4,7 @@ import { tmpdir } from 'node:os';
 import { createSilentSpinner } from '../utils/spinner.js';
 import { registerBot, fetchBot, updateBot } from './tdp.js';
 import { runAz } from '../utils/az.js';
+import { CliError } from '../utils/errors.js';
 import { logger } from '../utils/logger.js';
 
 export interface CreateBotOpts {
@@ -237,9 +238,15 @@ export async function discoverAzureBot(
       tenantId: account.tenantId,
       name: bot.name,
     };
-  } catch {
+  } catch (error) {
     spinner.stop();
-    return null;
+    const err = error as { stderr?: unknown; message?: unknown };
+    const stderr = typeof err.stderr === 'string' ? err.stderr.trim() : '';
+    const message =
+      stderr ||
+      (typeof err.message === 'string' && err.message) ||
+      'Unknown error discovering Azure bot.';
+    throw new CliError('API_ARM_ERROR', `Failed to discover Azure bot: ${message}`);
   }
 }
 
@@ -261,19 +268,15 @@ function subscriptionFromArmId(id: string): string | undefined {
 
 /** Fast path: assume Azure resource name == MicrosoftAppId. */
 async function findBotByName(botId: string): Promise<BotResource | null> {
-  try {
-    const results = await runAz<BotResource[]>([
-      'resource',
-      'list',
-      '--resource-type',
-      'Microsoft.BotService/botServices',
-      '--name',
-      botId,
-    ]);
-    return results[0] ?? null;
-  } catch {
-    return null;
-  }
+  const results = await runAz<BotResource[]>([
+    'resource',
+    'list',
+    '--resource-type',
+    'Microsoft.BotService/botServices',
+    '--name',
+    botId,
+  ]);
+  return results[0] ?? null;
 }
 
 /**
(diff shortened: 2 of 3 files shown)


packages/cli/src/commands/app/doctor.ts
```diff
@@ -217,7 +217,18 @@ async function checkBotRegistration(
...
+        azure = await discoverAzureBot(botId, silent);
+      } catch (e) {
+        discoveryFailed = true;
+        results.push({
+          category: cat,
+          label: 'Could not discover Azure bot',
+          status: 'fail',
+          detail: e instanceof Error ? e.message : undefined,
+        });
+      }
```
heyitsaamir, line 231: Hm, you don't need this now, right?

singhk97: it's needed because `disoverAzureBot`'s contract is changed and it now throws an error that `doctor` doesn't already handle.

singhk97:
I simulated a network error by overriding the windows DNS lookup for management.azure.com, and have it resolve to 127.0.0.1 (my local machine). Since there's no listener on that port a network calls to the endpoint will result in `no connection could be made because the target machine actively refused it` error.

Without the fix:

image

With the fix:

image

heyitsaamir (approved):

Issue #2091, opened 2024-10-07
[Dev support]: How to start a new turn with previous user message in the onUserSignInSuccess sign-in success callback


Benjiiim:
Is there a way to start a new turn with previous user message in the onUserSignInSuccess sign-in success callback?

I know how to get the message through the context.activity.text but I don't know what to do with it.

Thanks a lot

Benjiiim:
Might be related to #885

singhk97:
You could abstract away logic for the `onMessage` handler or handlers for `message` activity types and invoke them in the `onUserSignInSuccess` callback.

Benjiiim:
Thanks for your time @singhk97 
That can be a workaround for bots that rely on message activity types, but as you know better than me, one benefit of using Teams AI Library is to pass the user input to the AI Module when no activity handler are registered.
Is there a way to leverage this capability with the user input that triggered the sign-in flow?

lilyydu:
Hi @Benjiiim,

May I ask what you are planning to do with the turn and user input after the user is signed in? Another idea is registering an `afterTurn` handler for further processing. These handlers allow for the turn state to be modified/inspected after the turn is processed

Benjiiim:
I should have mentioned in my previous messages that I'm thinking about the [auto sign in](https://github.com/microsoft/teams-ai/blob/main/getting-started/CONCEPTS/USER-AUTH.md#auto-sign-in) feature, which means that the app will trigger the sign in flow even if the user is sending a message not related to authentication at all.

That can lead to weird exchanges such as:

User > I need help to do something
Bot (from the sign-in success callback) > You have been signed in. What can you do for you?
User > Well... ok but what about the help I've asked in my previous message?
Bot > I'm sorry, I don't know what you are referring to. How can I help?

That's why I'm looking for a way to have the user message processed by the app after a successful sign-in.

Does that make more sense?

lilyydu:
Yup, that makes more sense, thank you for clarifying

Unfortunately there is no straightforward way to begin a new turn during this callback. For auto-sign in, this scenario is primarily useful if the user should be signed in (by default) before they attempt to interact with the bot. I would suggest only following this route if that is the desired scenario

The only workaround I can think of as of now is to register some form of activity handler within the callback (though I know you mentioned you wanted to pipe this directly into the AI class)

Benjiiim:
[...]

Issue #885, opened 2023-11-24
[Feature Request]: simplify bot authentication dev experience


blackchoey:
**Describe your user scenario**

Currently, when adding authentication for bot, developers need to provide callback when sign-in success. The callback need to inspect the previous message stored in state and determine what to do against the message. This does not take the advantage of the message handlers registered to the `Application` class.

**Describe the solution you'd like**

Ideally, the SDK should continue processing the previous message after sign-in, without asking user to provide a callback function to process the messages.

Benjiiim:
May I ask if you have some news to share on this one?

On our app, I'm using the sign-in success callback for two things:
- Extract data from the JWT token to store it in the user state for future use (eg: name)
- Use the token to identify the user and know if he is allowed to use the app or not. If he is not allowed, we call the AuthenticationManager<ApplicationTurnState>.signOutUser() method.

Having a way to easily cancel the sign-in process or to create a new turn with the previous user message would be great improvements.

Benjiiim:
Hello @blackchoey
I'm wondering if you have update to share on this one? Should we expect a removal of this callback at some point? Should we expect improvements?
Thanks a lot

blackchoey:
@singhk97 can you help on Benjiiim's question?

singhk97:
I understand that this flow is not ideal but currently there are no plans to improve or change the success callback. We might decide to come back to it in the future.

Note that the `onUserSignInSuccess` would only be invoked if the sign in flow was initiated by a user message and not when done by a message extension or adaptive card. In those cases the original activity will be replayed by the Teams client.

Benjiiim:
A workaround has been discussed on #2091 with @corinagum

corinagum:
I discussed this with the team - this is one of those scenarios that, to solve the problem, we have to get opinionated about how users should handle their data. This is a gray area that we generally try to avoid. 

Instead, I think a good middle-ground would be to have documentation that clearly outlines different methods of handling this scenario, using in part the code linked above. I am working on adding this to our docs now.

MSFTRickyCastaneda:
Workaround provided according to chat, closing. Please reopen issue if still applicable

PR #2882, opened 2026-06-12
docs: add 'Resuming Pending Messages After Sign-In' guide

corinagum:
## Summary

Adds documentation for handling the original user message after an auto sign-in flow completes. This is a common scenario where the bot's sign-in flow consumes the user's initial message, causing a confusing UX.

The new section shows a pattern for storing the pending message before initiating sign-in, then replaying it in the `signin` event handler.

## Changes

- **Template**: Added `## Resuming Pending Messages After Sign-In` section to the user-authentication in-depth guide (before "Handling Sign-In Failures")
- **TypeScript fragment**: `pendingMessages` Map + `app.event('signin')` pattern
- **Python fragment**: `pending_messages` dict + `@app.event("sign_in")` with `SignInOptions`
- **C# fragment**: `ConcurrentDictionary` + `OAuthFlow.OnSignInComplete` (with `next/core` disclaimer)

## Context

#885 — workaround discussed in #2091. All code patterns verified against current SDK source.

### DIFF
--- a/teams.md/src/components/include/in-depth-guides/user-authentication/csharp.incl.md
+++ b/teams.md/src/components/include/in-depth-guides/user-authentication/csharp.incl.md
@@ -100,9 +100,9 @@ using System.Collections.Concurrent;
 var pendingMessages = new ConcurrentDictionary<string, (string Text, object Activity)>();
 
 // Get the pre-registered OAuth flow
-OAuthFlow auth = bot.GetOAuthFlow("graph");
+OAuthFlow auth = teams.GetOAuthFlow("graph");
 
-bot.OnMessage(async (context, cancellationToken) =>
+teams.OnMessage(async (context, cancellationToken) =>
 {
     // SignInAsync returns null if SSO was initiated (result arrives via OnSignInComplete)
     string? token = await auth.SignInAsync(context, cancellationToken);
--- a/teams.md/src/components/include/in-depth-guides/user-authentication/csharp.incl.md
+++ b/teams.md/src/components/include/in-depth-guides/user-authentication/csharp.incl.md
@@ -88,6 +88,53 @@ teams.OnMessage("/signout", async (context, cancellationToken) =>
     await context.Send("you have been signed out!", cancellationToken);
 });
 ```
+<!-- pending-messages -->
+
+:::note
+The C# OAuth APIs shown below (`OAuthFlow`, `SignInAsync`, `OnSignInComplete`) are available in the [`Microsoft.Teams.Apps`](https://www.nuget.org/packages/Microsoft.Teams.Apps) core package (2.1+ preview).
+:::
+
+```cs
+using System.Collections.Concurrent;
+
+var pendingMessages = new ConcurrentDictionary<string, (string Text, object Activity)>();
+
+// Get the pre-registered OAuth flow
+OAuthFlow auth = bot.GetOAuthFlow("graph");
+
+bot.OnMessage(async (context, cancellationToken) =>
+{
+    // SignInAsync returns null if SSO was initiated (result arrives via OnSignInComplete)
+    string? token = await auth.SignInAsync(context, cancellationToken);
+
+    if (token is null)
+    {
+        // Sign-in initiated — store the original message
+        var userId = context.Activity.From?.Id ?? string.Empty;
+        pendingMessages[userId] = (context.Activity.Text ?? string.Empty, context.Activity);
+        return;
+    }
+
+    // User is already signed in — process normally
+    await ProcessMessage(context.Activity.Text, context, cancellationToken);
+});
+
+auth.OnSignInComplete(async (context, tokenResponse, cancellationToken) =>
+{
+    var userId = context.Activity.From?.Id ?? string.Empty;
+
+    if (pendingMessages.TryRemove(userId, out var pending))
+    {
+        await context.SendActivityAsync("Successfully signed in! Processing your original request...", cancellationToken);
+        await ProcessMessage(pending.Text, context, cancellationToken);
+    }
+    else
+    {
+        await context.SendActivityAsync("You are now signed in!", cancellationToken);
+    }
+});
+```
+
 <!-- signin-failure -->
 
 ```cs
--- a/teams.md/src/components/include/in-depth-guides/user-authentication/python.incl.md
+++ b/teams.md/src/components/include/in-depth-guides/user-authentication/python.incl.md
@@ -99,16 +99,17 @@ pending_messages: dict[str, dict] = {}
 
 @app.on_message
 async def handle_message(ctx: ActivityContext[MessageActivity]):
-    if not ctx.is_signed_in:
-        # Store the original message before initiating sign-in
+    # sign_in() returns the token if already signed in, or None if OAuth card was sent
+    token = await ctx.sign_in(SignInOptions(
+        oauth_card_text="To help with that, I need to sign you in first."
+    ))
+
+    if token is None:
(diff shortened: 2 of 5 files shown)


teams.md/src/pages/templates/in-depth-guides/user-authentication.mdx
```diff
@@ -91,6 +91,18 @@ You can signout by calling the `signout` method, this will remove the token from
 
 <LanguageInclude section="signing-out" />
 
+## Resuming Pending Messages After Sign-In
+
+When using auto sign-in (calling `signin` inside a message handler), the user's original message is consumed by the sign-in flow. Once sign-in completes, your app receives a `signin` event — but the original message text is no longer available in that context.
+
+A common pattern is to store the original message before initiating sign-in, then retrieve and process it in the `signin` event handler. This avoids confusing exchanges where the bot ignores what the user originally asked.
```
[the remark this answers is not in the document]
corinagum: fixed

teams.md/src/components/include/in-depth-guides/user-authentication/csharp.incl.md
```diff
@@ -88,6 +88,53 @@ teams.OnMessage("/signout", async (context, cancellationToken) =>
...
+```cs
+using System.Collections.Concurrent;
+
+var pendingMessages = new ConcurrentDictionary<string, (string Text, object Activity)>();
+
+// Get the pre-registered OAuth flow
+OAuthFlow auth = bot.GetOAuthFlow("graph");
+
+bot.OnMessage(async (context, cancellationToken) =>
+{
```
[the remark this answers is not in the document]
corinagum: fixed

teams.md/src/pages/templates/in-depth-guides/user-authentication.mdx
```diff
@@ -91,6 +91,18 @@ You can signout by calling the `signout` method, this will remove the token from
...
+
+<LanguageInclude section="pending-messages" />
+
+:::tip
+For production apps, consider using a persistent store (database, Redis, etc.) instead of an in-memory map so pending messages survive restarts.
+:::
```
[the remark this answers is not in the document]
corinagum: fixed

teams.md/src/pages/templates/in-depth-guides/user-authentication.mdx
```diff
@@ -91,6 +91,18 @@ You can signout by calling the `signout` method, this will remove the token from
 
 <LanguageInclude section="signing-out" />
 
+## Resuming Pending Messages After Sign-In
+
+When using auto sign-in (calling `signin` inside a message handler), the user's original message is consumed by the sign-in flow. Once sign-in completes, your app receives a `signin` event — but the original message text is no longer available in that context.
```
nickwalkmsft: - The message is not available in the context of the signin event handler, but we awaited the signin call. Should the message handler always use the signin event handler to do work with the message, or can they just continue after awaiting the signin call? (See below comment too)
- Rephrase "the message is consumed"; it's really just a matter of being in the wrong context.
- The form of `signin` is language-dependent

teams.md/src/pages/templates/in-depth-guides/user-authentication.mdx
nickwalkmsft: - I suggest being more prescriptive than "A common pattern is"; just tell people this is how to do it.
- Is it not simpler to wait in the original message handler for the token to be set, if it's needed for this particular interaction? (See above comment).

teams.md/src/pages/templates/in-depth-guides/user-authentication.mdx
```diff
@@ -91,6 +91,18 @@ You can signout by calling the `signout` method, this will remove the token from
...
+
+A common pattern is to store the original message before initiating sign-in, then retrieve and process it in the `signin` event handler. This avoids confusing exchanges where the bot ignores what the user originally asked.
+
+<LanguageInclude section="pending-messages" />
+
+:::tip
+For production apps, consider using a persistent store (database, Redis, etc.) instead of an in-memory map so pending messages survive restarts.
```
nickwalkmsft: For the future: is there broader guidance about agent implementations that use a persistent store?

teams.md/src/components/include/in-depth-guides/user-authentication/csharp.incl.md
```diff
@@ -88,6 +88,53 @@ teams.OnMessage("/signout", async (context, cancellationToken) =>
...
+    // SignInAsync returns null if SSO was initiated (result arrives via OnSignInComplete)
+    string? token = await auth.SignInAsync(context, cancellationToken);
```
nickwalkmsft, line 108: Having a hard time understanding the ergonomics of this method call - it's async and returns a token, but the token might be null, and there's also an event that fires?

nickwalkmsft (commented):
Loving the real-world scenario guidance.

teams.md/src/components/include/in-depth-guides/user-authentication/python.incl.md
```diff
@@ -86,6 +86,46 @@ async def handle_signout_message(ctx: ActivityContext[MessageActivity]):
...
+            "text": ctx.activity.text,
+            "activity": ctx.activity,
+        }
+
+        await ctx.sign_in(SignInOptions(
+            oauth_card_text="To help with that, I need to sign you in first."
+        ))
+        return
+
+    # User is already signed in — process normally
```
heyitsaamir: can we avoid using is_signed_in? i think we need to start deprecating it. (more context in teams)

teams.md/src/pages/templates/in-depth-guides/user-authentication.mdx
corinagum: If `signin()` returns a token, (user was already signed in), you can continue processing in the same handler without an event. The pending-message is only needed when `signin()` returns undefined/None (OAuth card was sent) because the flow completes on a separate turn. I've updated the snippets to show this: check return value and only store/defer if it's null.

teams.md/src/pages/templates/in-depth-guides/user-authentication.mdx
corinagum: You can't, unfortunatley. When the OAuth card is sent, the current turn ends and the HTTP response is returned to Teams. The sign-in completion arrives as a new invoke (`signin/tokenExchange` or `signin/verifyState`) on a separate request. There's no way to block the original handler waiting for sign-in to finish.

teams.md/src/pages/templates/in-depth-guides/user-authentication.mdx
corinagum: Not at the moment. In the past we've actively avoided prescribing how stores are implemented / what tech to use, not sure why. For now, the tip is just to point out that in-memory won't survive restarts.

teams.md/src/components/include/in-depth-guides/user-authentication/csharp.incl.md
corinagum: `SigninAsync` tries silent token retrieval first. If a cached token exists, it returns it immediately (no event fires). If not, it sends the OAuth card and returns `null`. The result arrives later on `OnSignInComplete` when the client sends the token exchange invoke.

teams.md/src/components/include/in-depth-guides/user-authentication/python.incl.md
corinagum: Fixed

heyitsaamir (approved):

PR #2904, opened 2026-06-26
Fix CLI template issues and remove AI/MCP/MCPClient templates

heyitsaamir:
## Summary

Review of the CLI scaffolding templates surfaced several issues. This PR removes three problematic templates and fixes the rest.

### Removed templates
- **`typescript/ai`** — its `AGENTS.md` linked to a dead URL (`microsoft/teams.ts/main/AGENTS_TEMPLATE.md` returns 404).
- **`typescript/mcp`** — its `package.json.hbs` `inspect` script contained a literal newline, making the generated `package.json` invalid JSON (every npm command in a scaffolded project would break).
- **`typescript/mcpclient`** — removed alongside the MCP set.

### Fixes to remaining templates
- **graph (TS)** — declared missing `@microsoft/teams.api` dependency (imported for `MessageActivity`).
- **tab (TS)** — declared missing `@microsoft/teams-js` and `@microsoft/teams.common` dependencies; renamed leftover internal `@tests/tab` logger names to `tab`; awaited the `reply()` call in the echo handler.
- **graph (Python)** — fixed a possibly-unbound `me` (now initialized to `None` before the `if graph:` block) so pyright and runtime don't fail when `user_graph` is `None`; aligned the `dotenv` pin with the echo template.
- **Python READMEs** — removed reference to a nonexistent `apppackage/` directory.
- **menu-loop test** — updated the `listTemplates` mock to `['echo', 'graph', 'tab']`.

Template names are discovered dynamically from the filesystem via `listTemplates()`, so removing the directories is sufficient — no command code references them by name.

### Validation
- All remaining `package.json.hbs` files are valid JSON; every import is now a declared dependency.
- `npm run build` passes.
- All 206 tests pass.


corinagum (approved):

nickwalkmsft (approved):
@heyitsaamir @umangsehgal Where are templates on the roadmap? Would be good to get parity, and I'll also (eventually) be looking at getting an AGENTS.md in all of them.

PR #2934, opened 2026-07-23
docs: HTML Widgets in-depth guide

corinagum:
**Do not merge: putting on hold per request of HTML Widgets PM.**

## Summary

Adds documentation for the **HTML Widgets** feature (in-depth guide) covering all three SDKs (TypeScript, Python, .NET). HTML Widgets let an app render rich, interactive HTML inside a Teams message via an MCP Apps (ext-apps) surface.

## Pages added

Under **In-Depth Guides > HTML Widgets**:

- **Overview** (`README.mdx`) - what HTML Widgets are, how they travel inside an `extendedmarkdown` message, the payload shape, and a "What the SDK provides" helper index.
- **Sending Widgets** - building a widget message/markdown with the build helpers.
- **Handling Tool Calls** - the typed `htmlwidget/calltool` handler and typed response shape.
- **Security Policy** - `securityPolicy`, the validate/inject helpers, and their limits.
- **Updating Model Context** - parsing a widget's `ui/update-model-context` request off an incoming message.

## Notes

- Every page carries a **Preview** admonition (the feature and its API surface are still in preview).
- The Security Policy page includes a **"Not a security boundary"** note clarifying the helpers are development-time aids, not a replacement for the client's runtime enforcement or your own HTML review.


teams.md/src/pages/templates/in-depth-guides/html-widgets/README.mdx
```diff
@@ -0,0 +1,71 @@
...
+
+:::warning Preview
+HTML Widgets are in preview and the API surface may change.
+<LanguageInclude content={{"typescript": "Every widget type and helper is annotated `@experimental` in its JSDoc (diagnostic `ExperimentalTeamsHtmlWidget`) to mark it as preview.", "python": "Every widget type and helper is decorated with `@experimental` (`ExperimentalTeamsHtmlWidget`), which emits a runtime preview warning when it is used.", "csharp": "Every widget type and helper is guarded by the `ExperimentalTeamsHtmlWidget` compiler diagnostic, so you opt in explicitly to use them."}} />
+:::
+
+## How it works
+
+A widget travels inside a normal message.
+The app sends a message with `textFormat: 'extendedmarkdown'` whose text contains a fenced `html-widget` code block holding a JSON payload:
```
nickwalkmsft, line 23: Is there any functional reason a developer needs to know this, or is it just insight?

(Insight for its own sake is great, keep it regardless; I mostly ask because I'm wondering if it's worth putting a note in the extendedmarkdown docs that says "this feature also is the foundation for HTML widgets" and linking to widgets documentation)

teams.md/src/pages/templates/in-depth-guides/html-widgets/README.mdx
```diff
@@ -0,0 +1,71 @@
...
+---
+
+# HTML Widgets
+
+HTML Widgets let your app render rich, interactive HTML inside a Teams message.
+A widget is an [MCP Apps (ext-apps)](https://github.com/modelcontextprotocol/ext-apps) surface that the Teams client renders in a sandboxed iframe.
+The widget can call tools on your app, send messages back, request display mode changes, open links, and contribute content to the model context for future turns.
+
+:::warning Preview
+HTML Widgets are in preview and the API surface may change.
```
nickwalkmsft, line 16: Just taking this opportunity to mention - right now in the Learn docs, the concept of "preview" is used specifically to mean that use of the developer preview manifest schema is required, and that clients must have the _Public preview_ box checked. How closely does that align with what you mean here, and the strategy around preview/experimental features in the SDK in general? When we say it here, does it mean that developers have to do do those things to access the feature?

(Also, what's the timing on this, relative to `extendedmarkdown` actually hitting public preview?)

teams.md/src/pages/templates/in-depth-guides/html-widgets/sending-widgets.mdx
```diff
@@ -0,0 +1,34 @@
...
+## Adding text before and after
+
+You often want to introduce a widget with a sentence, or add a note after it.
+The build helpers accept options for text to place before and after the widget block.
+
+<LanguageInclude section="before-after-code" />
+
+## Building the markdown directly
+
+When you need full control over the outgoing activity, build just the markdown string and send it yourself.
```
nickwalkmsft: From the example, I was unable to see the utility of doing this relative to the Message version above.

teams.md/src/pages/templates/in-depth-guides/html-widgets/handling-tool-calls.mdx
```diff
@@ -0,0 +1,42 @@
...
+## The flow
+
+1. Your app sends a widget whose HTML calls `tools/call` over the MCP Apps protocol.
+2. The Teams client forwards that call to your app as an `htmlwidget/calltool` invoke.
+3. Your handler inspects the tool name and arguments, does its work, and returns a result.
+4. The Teams client delivers the result back to the widget, which re-renders.
+
+## Registering a handler
+
+Register a handler <LanguageInclude content={{"typescript": "for the `widget.callTool` event", "python": "with the `@app.on_widget_call_tool` decorator", "csharp": "with `teams.OnWidgetCallTool`"}} />. The activity value is typed, so you get the tool <LanguageInclude content={{"typescript": "`name` and `arguments`", "python": "`name` and `arguments`", "csharp": "`Name` and `Arguments`"}} /> directly.
```
nickwalkmsft, line 26: Another tangential question - are we pretty consistent on the usage of `teams` as the variable name for the main app object in C#?

teams.md/src/pages/templates/in-depth-guides/html-widgets/handling-tool-calls.mdx
```diff
@@ -0,0 +1,42 @@
...
+
+<LanguageInclude section="handler-code" />
```
nickwalkmsft, line 28: It was when I got here that I got wondering how familiar developers need to be with MCP and MCP apps as a concept to be comfortable use this, and how much this should explain that this is a specific protocol they're engaging with here. The shape of what I see here appears to be very intertwined with that protocol, as opposed to, say, our concept of _functions_ for tab apps, which is just RPC. Even the simple fact that a response is expected, let alone of a specific shape.

teams.md/src/pages/templates/in-depth-guides/html-widgets/handling-tool-calls.mdx
```diff
@@ -0,0 +1,42 @@
...
+A tool call result carries a list of content blocks, optional structured content the widget can render from, and an `isError` flag.
+Teams currently renders `text` content; the other MCP content types (`image`, `audio`, `resource`) are defined for forward compatibility.
+
+<LanguageInclude section="response-intro" />
+
+<LanguageInclude section="response-code" />
+
+## Throttling in large contexts
+
+In large or busy contexts the Teams platform may throttle widget tool calls, so your app can receive fewer `htmlwidget/calltool` invokes than the widget issued. This throttling happens on the platform before the invoke reaches your app, so there is no error or `429` status for your handler to catch and nothing for the SDK to retry. Design your tool handlers so that a dropped call is not fatal: keep them idempotent, and let the widget reflect its own state rather than assuming every call reaches your app.
```
nickwalkmsft, line 41: Not sure what _in large or busy contexts_ means; the key insight that should come through is whether that is something that's in the developer's control. I'm not sure if there's an implicit "don't make noisy widgets for your own good" here, or if it's more just "fyi sometimes it doesn't work lol".

What's the user experience when this happens?

teams.md/src/pages/templates/in-depth-guides/html-widgets/security-policy.mdx
```diff
@@ -0,0 +1,42 @@
...
+The browser's CSP enforcement is the real protection.
+Use it to catch a missing `connectDomains` or `resourceDomains` entry before you ship, so the widget does not silently fail to load a font or reach an API.
+
+Call <LanguageInclude content={{"typescript": "`validateSecurityPolicy`", "python": "`validate_security_policy`", "csharp": "`HtmlWidgetHelpers.ValidateSecurityPolicy`"}} /> with your HTML and the policy you intend to send. It returns <LanguageInclude content={{"typescript": "an array", "python": "a list", "csharp": "a list"}} /> of warnings; an empty <LanguageInclude content={{"typescript": "array", "python": "list", "csharp": "list"}} /> means every external reference is covered.
+
+<LanguageInclude section="validate-code" />
+
+Each warning tells you the offending URL, where it appeared in the HTML, and which policy field it belongs in, so you can correct the policy before sending.
+
+## Injecting the protocol script
```
nickwalkmsft: From the description, my first impression is that:

- This seems like a general widgets thing, not a security thing, and should be explained in "Sending Widgets" or somewhere more general?
- This seems like something that could/should be fully automatic?

teams.md/src/pages/templates/in-depth-guides/html-widgets/security-policy.mdx
```diff
@@ -0,0 +1,42 @@
+---
+sidebar_position: 4
+sidebar_label: 'Security Policy'
+title: 'Security Policy'
+---
+
+# Security Policy
+
+A widget's `securityPolicy` declares which origins the widget is allowed to reach for network requests, static resources, nested iframes, and base URIs.
```
nickwalkmsft, line 9: Needs brief coverage for the basic idea of constructing and attaching a security policy? This goes back to the question of "how much do we expect developers to already know about MCP apps, vs. them just wanting to use HTML widgets and MCP apps being an implementation detail".

teams.md/src/pages/templates/in-depth-guides/html-widgets/security-policy.mdx
```diff
@@ -0,0 +1,42 @@
...
+## Validating a policy
+
+`validateSecurityPolicy` scans your HTML for external references and returns a warning for any origin that is not covered by the appropriate policy field.
+This is a development-time audit tool, not a security boundary.
+The browser's CSP enforcement is the real protection.
+Use it to catch a missing `connectDomains` or `resourceDomains` entry before you ship, so the widget does not silently fail to load a font or reach an API.
+
+Call <LanguageInclude content={{"typescript": "`validateSecurityPolicy`", "python": "`validate_security_policy`", "csharp": "`HtmlWidgetHelpers.ValidateSecurityPolicy`"}} /> with your HTML and the policy you intend to send. It returns <LanguageInclude content={{"typescript": "an array", "python": "a list", "csharp": "a list"}} /> of warnings; an empty <LanguageInclude content={{"typescript": "array", "python": "list", "csharp": "list"}} /> means every external reference is covered.
+
+<LanguageInclude section="validate-code" />
```
nickwalkmsft: The narrative sort of addition at the bottom is confusing when all other snippets are direct examples of code that the reader could ship in prod.

Suggestion for this one is:

- Make this more production-ready by wrapping the validation warning generation in something that looks like an obvious debug flag.
- Remove the code from the bottom part and make it a narrative comment: "In development, this would warn that domain xyz is needed in this scenario, so that you could go modify the policy."

teams.md/src/pages/templates/in-depth-guides/html-widgets/updating-model-context.mdx
```diff
@@ -0,0 +1,47 @@
+---
+sidebar_position: 5
+sidebar_label: 'Updating Model Context'
+title: 'Updating Model Context'
+---
+
+# Updating Model Context
```
nickwalkmsft, line 7: I wasn't able to follow this article, but maybe it's because I don't understand Action.Submit yet.

Is this entirely about updating model context, or is that just the primary intended use case for this mechanism? FYI an interesting aspect of the platform docs right now is that there's a pretty hard separation between "here's a feature" and "here's how to use this feature with an LLM-based bot". In most cases the latter honestly doesn't even exist yet. I'm going to work on unifying those concerns (i.e. baking in an assumption that people are making LLM-powered bots) but it's going to take a while.

nickwalkmsft (commented):
Who is the feature PM for this? Widgets in general aren't on the docs backlog.

Sorry, forgot my general sentiment - this is great! Most of these comments are questions.

teams.md/src/pages/templates/in-depth-guides/html-widgets/README.mdx
corinagum: "Preview" - no 'public', is generally (in my mind at least) pointing out that this feature may not be available to you due to being in preview. No commitment either way about whether it needs public preview. API surface may change is further indicator that it's more volatile than public preview. 

"Public preview" mentioned specifically would be the indicator that it's in public preview and needs the opt-in. Going forward, we can try to include details that opting in via manifest is necessary, but so far I've found it hasn't been a problem.

teams.md/src/pages/templates/in-depth-guides/html-widgets/handling-tool-calls.mdx
corinagum: No, it looks like there's approximately even split between `app` and `teams`. This would be a broader unification effort, and PSA to the wider team.

teams.md/src/pages/templates/in-depth-guides/html-widgets/handling-tool-calls.mdx
corinagum: Yeah, HTML Widgets in particular is sticking fairly closely to the MCP-UI protocol, with some very minor differences that surface more internally than anything. I would say the biggest that is visible to the user from the Teams SDK perspective is that 'pip' is not a supported screen mode, but I made sure that was documented in the code.

teams.md/src/pages/templates/in-depth-guides/html-widgets/handling-tool-calls.mdx
corinagum: In extremely large chats (the hard number has changed multiple times so I have not provided one), it is possible widgets will get throttled. This is a service-side restriction and, while there is ongoing discussion about it, this is the current guidance. If in a large chat (50++ people), everyone is pressing a button in the widget at the same time, things will get throttled silently; there's no way to inform the user, and no 429 to put a condition on within the code. From the SDK perspective this is out of our control, but at the very least we can notify developers to prepare for it.

teams.md/src/pages/templates/in-depth-guides/html-widgets/security-policy.mdx
corinagum: Will move out of security policy. Yes, it is automatically applied. `injectWidgetProtocol` is the manual method.

teams.md/src/pages/templates/in-depth-guides/html-widgets/README.mdx
corinagum: Mostly just insight. They will need to set `extendedmarkdown`, so that aspect is necessary information.

teams.md/src/pages/templates/in-depth-guides/html-widgets/handling-tool-calls.mdx
nickwalkmsft: FYI as I work on docs I'm going to try to standardize on `teams`. I'd love to align with Py and TS but in .NET `app` is the conventional name for the result of `builder.Build()`.

PR #2942, opened 2026-07-31
docs: add Agent 365 in-depth guides

heyitsaamir:
## Summary

Adds the In-Depth Guide for Agent 365 support arriving in Teams SDK v2.1 (C#) and v2.1 preview (TypeScript and Python). The guide gives SDK users one path from the Agent 365 identity model through setup, messaging, lifecycle events, and governance-aware observability without duplicating Microsoft Learn’s tenant and product setup guidance.

This is a draft while the related SDK surfaces and observability integrations finish landing across the three implementations.

## Decisions

- **Identity terminology follows the upcoming SDK APIs.** The blueprint is the reusable runtime, the Agentic App Instance is its deployed identity, and the Agentic User is the conversational sender. `AgenticIdentity` is described as the operation scope rather than another identity type.
- **Shared narrative, language-specific code.** Templates keep concepts and guidance consistent while includes show the actual C#, TypeScript, and Python APIs, release status, token acquisition, lifecycle handlers, and proactive identity scoping.
- **Learn owns platform setup.** These pages explain what Teams SDK contributes and link to Microsoft Learn for Agent 365 onboarding, Microsoft OpenTelemetry Distro, Defender, and Purview configuration.
- **Observability layers stay distinct.** Teams SDK telemetry, Agent 365 baggage, and exporter configuration are presented separately so baggage is not mistaken for a complete governance pipeline.

## Deferred

Final API naming and preview details may be adjusted before this leaves draft as the corresponding language feature branches merge.


teams.md/src/pages/templates/in-depth-guides/agent-365/observability-and-governance.mdx
```diff
@@ -0,0 +1,61 @@
+---
+title: Observability and governance
+sidebar_position: 5
+summary: Connect Teams SDK telemetry to Agent 365 monitoring and governance while protecting sensitive data.
+---
+
+# Observability and governance
+
+Agent 365 observability has three layers:
```
MehakBindra, line 9: C# already has a Otel page under Observability . should we add something similar for PY and TS ? And link to it here ? As those concepts are related but don't belong here

heyitsaamir: Yup. Will add it separately

corinagum (approved):

teams.md/src/pages/templates/in-depth-guides/agent-365/README.mdx
```diff
@@ -0,0 +1,22 @@
+---
+title: Agentic Identity
+sidebar_position: 2
+summary: Build Teams agents with Agent 365 identities, lifecycle events, observability, and governance.
+suppressLanguageIncludeWarning: true
+---
+
+# Agentic Identity
+
+[Microsoft Agent 365](https://learn.microsoft.com/microsoft-agent-365/) gives your agent a governed identity in Microsoft 365 through an agent identity blueprint. Teams SDK handles the Activity Protocol and identity-aware messaging that your agent needs in Teams.
```
nickwalkmsft, line 10: How much of all of this is better covered by the Agent ID docs, as opposed to Agent 365? I honestly don't know. https://learn.microsoft.com/en-us/entra/agent-id/. It has its own AI-guided setup and everything, https://learn.microsoft.com/en-us/entra/agent-id/agent-id-ai-guided-setup

teams.md/src/pages/templates/in-depth-guides/agent-365/README.mdx
```diff
@@ -0,0 +1,22 @@
...
+
+<LanguageInclude section="availability-note" />
```
nickwalkmsft, line 12: Outdated now

teams.md/src/pages/templates/in-depth-guides/agent-365/README.mdx
```diff
@@ -0,0 +1,22 @@
...
+
+With Teams SDK, your app can:
```
nickwalkmsft, line 14: See top comment - what scenario/persona are we addressing here? I'm not saying it's not a valid one, just that it's not clear what it is (what they know, what they already have, what they want to - or might want to - achieve).

teams.md/src/pages/templates/in-depth-guides/agent-365/concepts-and-benefits.mdx
```diff
@@ -0,0 +1,55 @@
...
+  B --> I2[Agentic App Instance B]
+  I1 --> U1[Agentic User A]
+  I2 --> U2[Agentic User B]
+```
+
+One blueprint can power many agents. Your code uses the Agentic User and app instance IDs to load the correct configuration, state, and permissions for each agent.
+
+For a deeper identity architecture overview, see [Plan an agent identity architecture](https://learn.microsoft.com/entra/agent-id/how-to-plan-agent-identity-architecture).
+
+## How Agentic Users Compare to a Standard Bot
```
nickwalkmsft, line 35: How do they compare _within the scope of Teams_? For both end users and developers. This is focused on implementation, but we need to explain the difference in experience, capabilities and mental model first, for both developers and end users (because the former needs to understand what they're building for the latter).

teams.md/src/pages/templates/in-depth-guides/agent-365/setup.mdx
```diff
@@ -0,0 +1,45 @@
+---
+title: Set up Agentic Users
+sidebar_position: 2
+summary: Set up Agent 365, then configure the identity values required by Teams SDK.
+suppressLanguageIncludeWarning: true
+---
+
+import LangLink from '@site/src/components/LangLink';
+
+# Set up Agentic Users
```
nickwalkmsft, line 10: I can't tell where this is starting from - the linked quickstart is "Connect an existing agent", but what does that refer to? A Teams SDK agent that the user already has?

nickwalkmsft (commented):
Biggest overarching question is "who are we trying to address?" That question is largely for @umangsehgal and is larger than the scope of work here, but it does apply here too.

In particular: what does our audience know, what do they already have, and what are (or maybe should) they trying to do? Are we trying to help developers choose between a traditional bot and agent ID? We don't necessarily have to address every concern of every persona, but we should be explicit about which ones we _are_ addressing and about assumptions we're making about them.

For a first swing at docs on this topic, I think we should actually omit explanations of a lot of agent ID concepts and benefits, assume that the reader knows what they are, is likely using them already and doesn't need to make a choice, and focus on implementation and experiential concerns for them specific to Teams and Teams SDK. It's better to have a confident, solid set of docs for a "smaller" persona than you'd like than an unfocused set of docs that tries to address a larger audience but ends up confusing them because they're not sure who it's for. From there, we can explore additional articles to address other personas that need more information, and make sure that we address them properly as we go.

PR #2975, opened 2026-09-02
docs: update TS/PY auth docs for multi-connection oauth

lilyydu:
adds tabs across the TypeScript and Python user-authentication guide and updates examples to the new per-connection OAuth flow API

- user-authentication
- essentials/graph: 
- essentials/on-event: 
- migrations (botbuilder + v1)

### DIFF
--- a/teams.md/src/components/include/essentials/graph/python.incl.md
+++ b/teams.md/src/components/include/essentials/graph/python.incl.md
@@ -31,9 +31,15 @@ You also have access to the `app_graph` object in the activity handler. This is 
 
 <!-- user-graph-intro -->
 
-You can also access the graph using the user's token from within a message handler via the `user_graph` property.
+You can also access the graph using the user's token from within a message handler.
 
 <!-- user-graph-example -->
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+<Tabs groupId="python-sdk-version" defaultValue="core">
+<TabItem value="legacy" label="SDK 2.1 (Legacy)">
 
 ```python
 @app.on_message
@@ -45,6 +51,33 @@ async def handle_message(ctx: ActivityContext[MessageActivity]):
     print(f"User Job Title: {user.job_title}")
 ```
 Here, the "user_graph" object is a scoped graph client for the user that sent the message.
+
+</TabItem>
+<TabItem value="core" label="SDK 2.2 (current)" default>
+
+Build the client from an OAuth flow's token, so it's always scoped to the connection that owns it.
+
+```python
+from microsoft_teams.graph import get_graph_client
+
+graph = app.add_oauth_flow("graph")
+
+@app.on_message
+async def handle_message(ctx: ActivityContext[MessageActivity]):
+    token = await graph.sign_in(ctx)
+    if token is None:
+        return  # OAuth card sent — resumes on the callback turn
+
+    client = get_graph_client(token)
+    user = await client.me.get()
+    print(f"User ID: {user.id}")
+    print(f"User Display Name: {user.display_name}")
+    print(f"User Email: {user.mail}")
+    print(f"User Job Title: {user.job_title}")
+```
+
+</TabItem>
+</Tabs>
 
 <!-- advanced-sections -->
 
--- a/teams.md/src/components/include/essentials/graph/python.incl.md
+++ b/teams.md/src/components/include/essentials/graph/python.incl.md
@@ -39,7 +39,7 @@ import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 
 <Tabs groupId="python-sdk-version" defaultValue="core">
-<TabItem value="legacy" label="SDK 2.1 (Legacy)">
+<TabItem value="legacy" label="SDK 2.0 (Legacy)">
 
 ```python
 @app.on_message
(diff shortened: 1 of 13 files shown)


teams.md/src/pages/templates/in-depth-guides/user-authentication.mdx
```diff
@@ -49,19 +49,19 @@ Make sure you use the same name you used when creating the OAuth connection in t
 
 <Language language={['typescript', 'python']}>
 :::note
-In many templates, `graph` is the default name of the OAuth connection, but you can change that by supplying a different connection name in your app configuration.
+In many templates, `graph` is the default name of the OAuth connection, but you can register the connection under any name — it just has to match the OAuth connection configured on the Azure Bot.
```
nickwalkmsft, line 53: Have seen this in a couple PRs, so making sure it's not just me - `:::note`s don't render on the site.

nickwalkmsft (commented):
When it's convenient, could you draft a short (just a couple sentences) summary of what multi-auth _is_, in the form of a feature announcement?

Is this a breaking change for people upgrading from 2.1?

teams.md/src/components/include/essentials/graph/python.incl.md
```diff
@@ -46,6 +52,33 @@ async def handle_message(ctx: ActivityContext[MessageActivity]):
 ```
 Here, the "user_graph" object is a scoped graph client for the user that sent the message.
 
+</TabItem>
+<TabItem value="core" label="SDK 2.1 (current)" default>
+
+Build the client from an OAuth flow's token, so it's always scoped to the connection that owns it.
+
+```python
```
MehakBindra, line 60: for all languges the transition from app graph to user graph on this page happens very quietly, I think we should make the distinction clearer .

Also TS has many more sections, wondering if we can trim/standardize as part of this PR ?

teams.md/src/components/include/in-depth-guides/user-authentication/csharp.incl.md
```diff
@@ -253,6 +253,14 @@ teams.OnMessage("/signout github", async (context, cancellationToken) =>
   </TabItem>
 </Tabs>
 
+<!-- multiple-connections -->
```
MehakBindra, line 256: Can we also change C# to match PY and TS here ? or vice versa? Right now C# has the multiple connections section but they have already been introduced in signing in, seems weird 

Same for connection status , can we include in C# too please ?

teams.md/src/components/include/essentials/on-event/typescript.incl.md
```diff
@@ -42,10 +42,18 @@ app.event('error', ({ error }) => {
 When a user signs in using `OAuth` or `SSO`, use the graph api to fetch their profile and say hello.
 
 ```typescript
+import { Client as GraphClient } from '@microsoft/teams.graph';
```
MehakBindra, line 45: Should we make PY and TS have the same examples ? Right now they're different

corinagum (approved):
lgtm; agree with pre-existing comments :)

teams.md/src/components/include/essentials/graph/python.incl.md
lilyydu: looked into this — the differences are moreso architectural. TS ships its own client and tree-shakable endpoints packages, so the calls are different, and I hesitate to remove these here. If we want to do this, I think it should be a separate PR since it's leaning out of scope.

teams.md/src/components/include/essentials/on-event/typescript.incl.md
lilyydu: yess! good catch

teams.md/src/components/include/essentials/graph/typescript.incl.md
```diff
@@ -67,6 +73,34 @@ app.on('message', async ({ activity, userGraph }) => {
...
+
+const graph = app.addOAuthFlow('graph');
+
+// Gets details of the current user
+app.on('message', async (ctx) => {
+  const token = await graph.signIn(ctx);
+  if (!token) return; // OAuth card sent — resumes on the callback turn
+
+  const client = new GraphClient({ token: () => token }, { baseUrlRoot: app.graphBaseUrl });
+  const me = await client.call(endpoints.me.get);
```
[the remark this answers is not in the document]
lilyydu: false positive, this is correct. there's two tokens here and both usages are right

teams.md/src/components/include/in-depth-guides/user-authentication/typescript.incl.md
```diff
@@ -82,18 +160,102 @@ app.on('message', async ({ send, activity, signin }) => {
...
+```ts
+import { Client as GraphClient } from '@microsoft/teams.graph';
+import * as endpoints from '@microsoft/teams.graph-endpoints';
+
+app.message('/whoami', async (ctx) => {
+  const token = await graph.signIn(ctx);
+  if (!token) return; // OAuth card sent — resumes on the callback turn
+
+  const client = new GraphClient({ token: () => token }, { baseUrlRoot: app.graphBaseUrl });
+  const me = await client.call(endpoints.me.get);
```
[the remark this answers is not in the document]
lilyydu: same false positive as above^

corinagum (approved):

PR #2979, opened 2026-09-08
docs: add TypeScript and Python state guides

MehakBindra:
## Summary

- expand the state management guide to TypeScript and Python using their SDK 2.1 APIs
- move shared conversation/user scope and turn lifecycle behavior into the common template
- document language-specific setup, reads/writes, clearing, and durable storage configuration
- retain the C# guide as a language include and remove TypeScript/Python from the missing-pages manifest

## Validation

- generated all language-specific documentation
- built the Docusaurus site successfully

### DIFF
--- a/teams.md/src/components/include/in-depth-guides/state/python.incl.md
+++ b/teams.md/src/components/include/in-depth-guides/state/python.incl.md
@@ -31,11 +31,19 @@ async def handle_message(ctx: ActivityContext[MessageActivity]) -> None:
     await ctx.send(f"Hello, {name}. Message #{count}.")
 ```
 
+<!-- state-operations -->
+
+Use `key in scope` to check for a value, `del scope[key]` to remove one, and `clear()` to remove all values. Changes inside stored lists and dictionaries are detected automatically when the turn is saved.
+
 <!-- clear-example -->
 
 ```python
 await ctx.state.delete()
 ```
+
+<!-- distributed-intro -->
+
+For production or multi-instance deployments, implement the Teams SDK's `Storage[str, Any]` contract with a shared, durable backend and pass it through `StateOptions`. The state API used by handlers doesn't change:
 
 <!-- distributed-example -->
 
@@ -54,3 +62,7 @@ def create_app(durable_storage: Storage[str, Any]) -> App:
         )
     )
 ```
+
+<!-- distributed-details -->
+
+The SDK serializes each scope as a JSON string and replaces the complete scope on save, so concurrent turns use last-writer-wins semantics. Configure expiry, retries, and other storage-specific behavior on your storage implementation.
--- a/teams.md/src/components/include/in-depth-guides/state/python.incl.md
+++ b/teams.md/src/components/include/in-depth-guides/state/python.incl.md
@@ -1,10 +1,4 @@
-<!-- overview -->
-
-The Teams SDK provides opt-in, per-turn state for storing conversation and user data across activities. State is loaded before each activity handler runs, saved automatically after the turn, and exposed through `ctx.state`.
-
-<!-- setup -->
-
-Set the `App` option `state` to `True`. With no storage provider configured, state uses process-local `LocalStorage` and is lost when the process restarts.
+<!-- setup-example -->
 
 ```python
 from microsoft_teams.apps import App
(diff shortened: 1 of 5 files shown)


teams.md/src/components/include/in-depth-guides/state/typescript.incl.md
```diff
@@ -0,0 +1,74 @@
+<!-- overview -->
+
+The Teams SDK provides opt-in, per-turn state for storing conversation and user data across activities. State is loaded before each activity handler runs, saved automatically after the turn, and exposed through `ctx.state`.
+
+<!-- setup -->
+
+Set the `App` option `state` to `true`. With no storage provider configured, state uses process-local `LocalStorage` and is lost when the process restarts.
```
nickwalkmsft:
```suggestion
Set the `App` option `state` to `true`. With no storage provider configured, state uses an in-memory `Map<string, T>` and is lost when the process restarts.
```

I suggest something like this (and similar for Python), this reads like `LocalStorage` is a standard-library feature when that's not the case.

teams.md/src/pages/templates/in-depth-guides/state.mdx
```diff
@@ -2,79 +2,34 @@
...
-    // Write
-    context.State.ConversationState.Set("lastMessage", context.Activity.Text ?? string.Empty);
-    context.State.UserState.Set("messageCount",
-        (context.State.UserState.Get<int>("messageCount")) + 1);
+Conversation state is shared by everyone in the current conversation. User state is scoped to the current sender **within that conversation** and is unavailable when an activity has no usable sender ID.
 
-    // Read
-    string? last = context.State.ConversationState.Get<string>("lastMessage");
-    int count = context.State.UserState.Get<int>("messageCount");
+Only changed scopes are persisted. After the handler chain finishes, the SDK saves state and seals both scopes. Do not retain turn state for background work; accessing a sealed scope throws an error.
```
nickwalkmsft: Are these concepts that only apply to .NET? I'm not sure what they mean or how they would work in what I see provided in Python and TypeScript.

teams.md/src/components/include/in-depth-guides/state/python.incl.md
```diff
@@ -0,0 +1,72 @@
...
+
+```python
+await ctx.state.delete()
+```
+
+Values written after `delete()` are saved normally at the end of the current turn.
+
+<!-- distributed-state -->
+
+Process-local storage is intended for development only. For production or multi-instance deployments, pass a shared `Storage[str, Any]` implementation through `StateOptions`. The state API used by handlers does not change:
```
nickwalkmsft: Similar sentiment here as the comment about `LocalStorage`; this reads like it's something the user is expected to know about from a standard library. You don't necessarily need to document the whole interface, but at least call out that it's something they'll need to implement.

teams.md/src/components/include/in-depth-guides/state/python.incl.md
```diff
@@ -0,0 +1,72 @@
+<!-- overview -->
+
+The Teams SDK provides opt-in, per-turn state for storing conversation and user data across activities. State is loaded before each activity handler runs, saved automatically after the turn, and exposed through `ctx.state`.
```
corinagum: Seems like a lot of the explanations can be de-duped, lower maintenance burden. Could you do a pass?

corinagum: ( `<LanguageInclude content={} />` for inline)

teams.md/src/components/include/in-depth-guides/state/python.incl.md
```diff
@@ -0,0 +1,72 @@
...
+
+Set the `App` option `state` to `True`. With no storage provider configured, state uses process-local `LocalStorage` and is lost when the process restarts.
+
+```python
+from microsoft_teams.apps import App
+
+app = App(state=True)
+```
+
+State is disabled by default. When it is disabled, `ctx.state` is `None`.
```
lilyydu: I think it's worth flagging as a small note here - if they do `add_oauth_flow` - it will autoenable state

teams.md/src/pages/templates/in-depth-guides/state.mdx
MehakBindra: They work for all languages, let me try to make this clearer

teams.md/src/components/include/in-depth-guides/state/python.incl.md
MehakBindra: yes, fair point

corinagum (approved):

PR #2989, opened 2026-09-15
docs: document the two minute streaming limit and long-run handoff

corinagum:
## Why

The streaming docs documented only the happy path. It doesn't mention the Teams service's two minute limit on the lifetime of a `streamId`, what happens when an agent exceeds it, or what to do about an agent that needs longer. 

## What this adds

Three subsections under `## Streaming` in `essentials/sending-messages`:

- **The two minute streaming limit.** Documents the limit and the `403` / `ContentStreamNotAllowed` response, and links to the [streaming error codes](https://learn.microsoft.com/en-us/microsoftteams/platform/bots/streaming-ux#error-codes) table. That table is the only place this is documented publicly; notably it is absent from the constraints list at the top of the same page, which is likely why people keep hitting it by surprise. Also explains that exceeding the limit is **not** data loss in current SDKs: the stream is marked timed out, buffering continues, and close finalizes by updating the original activity in place.
- **Running longer than two minutes.** The recovery pattern: stream normally, then hand off before the window closes by finalizing the stream and switching to plain activity updates on the same message. No streaming window applies to plain edits, so the answer can keep growing as a single message with no extra messages in the thread and no repeat notifications.
- **Starting another streamed message.** Emitting after close opens a new streamed message with its own fresh window, which is for a genuinely separate answer rather than a way to extend one.

Plus a `:::warning` for the gotcha that the clock starts on the **first streamed chunk**, and an informative update counts as a chunk. An agent that posts "Thinking..." immediately and then works for three minutes has spent its entire window before emitting a single token.

**The C# sample is not at parity with TypeScript and Python, on purpose.** TS and Python set an id on the outgoing activity so the send routes through the update path, keeping everything in one message. C# cannot: `FinalizeResponseAsync` returns a bare `Task` and discards the finalized message id (`_streamId` is private), and `Context.SendAsync` has no id check so it always creates. The C# sample therefore finalizes, sends one follow-up message, and edits that, costing one extra message in the thread. The fragment states this limitation rather than implying parity. If we want true parity this looks like a real SDK gap worth its own issue.

**The C# sample tracks `handedOff` separately from `messageId`.** `SendActivityResponse.Id` is nullable, so reusing a null `messageId` as the "not yet handed off" flag would re-run the handoff branch on every later chunk and post a new message each time. There is a short paragraph in the fragment explaining why the two variables are separate, so the next person to touch it does not collapse them and reintroduce the bug.

**The C# section has no SDK 2.0 (Legacy) tab.** 2.0 source is not on `main` (`Libraries/` and `core/` contain no `.cs` files, the solution references only `src/*`, and `RELEASE.md` puts 2.0 on `release/v2.0`), so there is no surface to document. The intro line says so explicitly, because `groupId="csharp-sdk-version"` is sticky page-wide and a reader parked on the Legacy tab would otherwise hit an untabbed section with no explanation.

**Deliberately not asserted.** No specific throttle figure for plain updates; the one request per second number applies to the streaming path, not to edits, so the text says "a cadence measured in seconds". Nothing about whether Teams renders an "edited" marker on a previously streamed message. Neither is verified.

**Samples are not compile tested.** Every method name and signature was verified against SDK source (`teams.ts/packages/apps/src/http/http-stream.ts` and `activity-sender.ts`, `teams.py` `StreamerProtocol` / `activity_send.py` / `SentActivity`, `teams.net/src/Microsoft.Teams.Apps/TeamsStreamingWriter.cs` and `Context.cs`), but no compiler has been run against the snippets, and they reference a placeholder agent function by design.

## Validation

`npm run generate:docs` reports `0 template(s) have missing sections` across 68 templates for all three languages, matching the pre-change baseline, so the new `streaming-handoff-example` fragment resolves everywhere with no content gaps. `npm run build` succeeds. The two broken anchor warnings in `migrations/botbuilder/integration` are pre-existing and unrelated.


teams.md/src/pages/templates/essentials/sending-messages/README.mdx
```diff
@@ -32,11 +32,35 @@ You may also stream messages to the user which can be useful for long messages,
 <LanguageInclude section="streaming-example" />
 
 :::note
-Streaming is currently only supported in 1:1 conversations, not group chats or channels
+Streaming is currently only supported in 1:1 conversations, not group chats or channels. Teams also supports only one concurrent streaming response per chat at a time.
+
+Every inbound activity gets its own stream, and the SDK closes that stream at the end of the turn but does not serialize turns for you. If your agent can be re-prompted while it is still responding, serialize turns per conversation or finalize the in-flight stream before starting the next one.
```
nickwalkmsft: > If your agent can be re-prompted

This isn't really an "if", is it? I don't think it's really about _re-prompting_, it's about whether another `onmessage` might fire concurrently, which I would assume is always a possibility?

As far as I can tell, there's a general caution here that when you use streaming you introduce concurrency concerns with message handling that don't exist otherwise. Its OK if we want to punt on more complete guidance around that for now, but go ahead and just drop in a sentence about it if that's the case. It's better to leave the solution as an exercise for the reader than to not mention it as all.

teams.md/src/pages/templates/essentials/sending-messages/README.mdx
```diff
@@ -32,11 +32,35 @@ You may also stream messages to the user which can be useful for long messages,
...
 :::
 
 ![Animated image showing agent response text incrementally appearing in the chat window.](/screenshots/streaming-chat.gif)
 
+### The two minute streaming limit
+
+Teams keeps a streamed message open for two minutes, measured from its first chunk. After that the service rejects further chunks for that message with a `403` and the [error code](https://learn.microsoft.com/en-us/microsoftteams/platform/bots/streaming-ux#error-codes) `ContentStreamNotAllowed` (`Content stream finished due to exceeded streaming time.`). The window is scoped to the individual streamed message rather than to your agent, so [a later streamed message](#starting-another-streamed-message) opens a fresh one.
```
nickwalkmsft:
```suggestion
Teams keeps a message stream open for two minutes, measured from its first chunk. After that the service rejects further chunks for that message with a `403` and the [error code](https://learn.microsoft.com/en-us/microsoftteams/platform/bots/streaming-ux#error-codes) `ContentStreamNotAllowed` (`Content stream finished due to exceeded streaming time.`).
```

Last sentence was confusing, I dropped it

teams.md/src/pages/templates/essentials/sending-messages/README.mdx
```diff
@@ -32,11 +32,35 @@ You may also stream messages to the user which can be useful for long messages,
...
+
+However, hitting the limit does not mean the response is lost. The SDK marks the stream as timed out and stops pushing chunks, but it keeps buffering everything your agent produces, and <LanguageInclude content={{"typescript": "`stream.close()`", "python": "`ctx.stream.close()`", "csharp": "`writer.FinalizeResponseAsync()`"}} /> finalizes by updating the original message in place with the full buffered text. 
+
+:::warning
+The two minute clock starts on the **first streamed chunk**, and an informative update (<LanguageInclude content={{"typescript": "`stream.update()`", "python": "`ctx.stream.update()`", "csharp": "`writer.SendInformativeUpdateAsync()`"}} />) is a chunk. An agent that posts a status update immediately and then thinks for three minutes has spent its entire window before emitting a single token. When the slow work happens before you have any output, do that work first and open the stream once real content starts.
```
nickwalkmsft: Good guidance

teams.md/src/pages/templates/essentials/sending-messages/README.mdx
```diff
@@ -32,11 +32,35 @@ You may also stream messages to the user which can be useful for long messages,
...
+
+To outlast the window, hand off before it closes: stream normally, then finalize the stream and continue the response growing with plain activity updates. No streaming window applies to plain updates, so the answer can keep arriving for as long as your agent runs.
+
+<LanguageInclude section="streaming-handoff-example" />
+
+Plain updates are subject to ordinary activity throttling rather than the streaming path's allowance, so pace them at a cadence measured in seconds rather than sending one per token.
+
+### Starting another streamed message
+
+Calling <LanguageInclude content={{"typescript": "`stream.emit()`", "python": "`ctx.stream.emit()`", "csharp": "`writer.AppendResponseAsync()`"}} /> again after the stream has been closed opens a new streamed message on the same stream instance, with its own two minute window. Use this when your agent has a genuinely separate second answer to stream, rather than as a way to extend a single response.
```
nickwalkmsft: Wasn't able to follow this, "same stream instance" makes it sound like you can keep streaming the same message even after two minutes, but the latter half says this is talking about "a genuinely separate second answer to stream."

teams.md/src/pages/templates/essentials/sending-messages/README.mdx
```diff
@@ -32,11 +32,35 @@ You may also stream messages to the user which can be useful for long messages,
...
+The two minute clock starts on the **first streamed chunk**, and an informative update (<LanguageInclude content={{"typescript": "`stream.update()`", "python": "`ctx.stream.update()`", "csharp": "`writer.SendInformativeUpdateAsync()`"}} />) is a chunk. An agent that posts a status update immediately and then thinks for three minutes has spent its entire window before emitting a single token. When the slow work happens before you have any output, do that work first and open the stream once real content starts.
+:::
+
+### Running longer than two minutes
+
+To outlast the window, hand off before it closes: stream normally, then finalize the stream and continue the response growing with plain activity updates. No streaming window applies to plain updates, so the answer can keep arriving for as long as your agent runs.
+
+<LanguageInclude section="streaming-handoff-example" />
+
+Plain updates are subject to ordinary activity throttling rather than the streaming path's allowance, so pace them at a cadence measured in seconds rather than sending one per token.
```
nickwalkmsft: Now that we're mentioning activity updates (do we anywhere else?) the absence of guidance about why to stream instead is more apparent. Maybe drop in a quick sentence or two at the top that explains what streaming is for relative to updates (liveness UX, better latency I'm assuming, ~~informative updates~~ thinking indicators, stoppability, anything else?)

nickwalkmsft (approved):
I'm torn on whether the handoff guidance is too intricate and specialized to have right here with the basic feature explanation, but let's leave it for now, it's well-thought-out.

teams.md/src/pages/templates/essentials/sending-messages/README.mdx
corinagum: You're right. I'll reword so it says handlers can run concurrently.

teams.md/src/pages/templates/essentials/sending-messages/README.mdx
corinagum: Sorry, yeah it's confusing, I'll attempt a reword. The object in the handler gets reused, and a second message comes with its own timeout/id.

